PR #18, opened 2019-01-01
Assume/Assert Cell Support

cr1901:
## Scope
This PR adds the beginning of formal verification support to `nmigen`. With this PR, a user can now write statements of the following form:

```
m.d.comb += Assert(foo == bar)
m.d.sync += Assert(baz == 0)

with m.If(quux):
    m.d.comb += Assume(dummy == 1)
```

The first statement in particular will create `Assert`s/`Assume`s analogous to `yosys`'s "`properties`":
```
assert property (foo == 0)
```

A `yosys` `property` is syntactic sugar for:
```
always @(*) assert(foo == 0)
```

The third statement is analogous to a SystemVerilog immediate `assert`/`assume`.

The generated RTLIL can then be fed to `yosys`, which outputs an `smtv2` file. Using `yosys-smtbmc` or `symbiyosys`, one can now run a basic formal verification flow using `nmigen`!

## Limitations
* Something analogous to `initial assume/assert` not yet supported (easy to add initial support, but I don't know what the semantics should be if they are used within an `If/Else/Switch` context.
* You can only do immediate asserts (in SystemVerilog, `property` is for concurrent asserts, and AFAIK `yosys` does not support these).

### DIFF
--- a/nmigen/back/pysim.py
+++ b/nmigen/back/pysim.py
@@ -297,6 +297,12 @@ class _StatementCompiler(StatementVisitor):
             lhs(state, normalize(rhs(state), shape))
         return run
 
+    def on_Assert(self, stmt):
+        raise NotImplementedError("Asserts not yet implemented for Simulator backend.") # :nocov:
+
+    def on_Assume(self, stmt):
+        pass # :nocov:
+
     def on_Switch(self, stmt):
         test  = self.rrhs_compiler(stmt.test)
         cases = []
--- a/nmigen/hdl/ast.py
+++ b/nmigen/hdl/ast.py
@@ -12,8 +12,9 @@ __all__ = [
     "Value", "Const", "C", "Operator", "Mux", "Part", "Slice", "Cat", "Repl",
     "Array", "ArrayProxy",
     "Signal", "ClockSignal", "ResetSignal",
-    "Statement", "Assign", "Switch", "Delay", "Tick", "Passive",
-    "ValueKey", "ValueDict", "ValueSet", "SignalKey", "SignalDict", "SignalSet",
+    "Statement", "Assign", "Assert", "Assume", "Switch", "Delay", "Tick",
+    "Passive", "ValueKey", "ValueDict", "ValueSet", "SignalKey", "SignalDict",
+    "SignalSet",
 ]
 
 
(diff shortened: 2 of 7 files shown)


nmigen/back/pysim.py
```diff
@@ -291,6 +291,12 @@ def run(state):
             lhs(state, normalize(rhs(state), shape))
         return run
 
+    def on_Assert(self, stmt):
+        raise NotImplementedError("Asserts not yet implemented for Simulator backend.") # :nocov:
+
+    def on_Assume(self, stmt):
+        raise NotImplementedError("Assumes not yet implemented for Simulator backend.") # :nocov:
```
whitequark: Assumes should be ignored during simulation.

nmigen/hdl/xfrm.py
```diff
@@ -164,6 +176,14 @@ def on_value(self, value):
     def on_Assign(self, stmt):
         return Assign(self.on_value(stmt.lhs), self.on_value(stmt.rhs))
 
+    def on_Assert(self, stmt):
+        # Preserve _en, as the compiler state thinks _en is driven.
+        return Assert(self.on_value(stmt.test), _en=stmt._en)
+
+    def on_Assume(self, stmt):
+        # Preserve _en, as the compiler state thinks _en is driven.
+        return Assume(self.on_value(stmt.test), _en=stmt_en)
```
whitequark: Typo.

nmigen/hdl/xfrm.py
```diff
@@ -164,6 +176,14 @@ def on_value(self, value):
     def on_Assign(self, stmt):
         return Assign(self.on_value(stmt.lhs), self.on_value(stmt.rhs))
 
+    def on_Assert(self, stmt):
+        # Preserve _en, as the compiler state thinks _en is driven.
```
whitequark: This comment is incorrect; `_en` *is* in fact driven. Besides, preserving parts of statements during transforms doesn't really need additional comments.

nmigen/hdl/xfrm.py
```diff
@@ -328,6 +354,12 @@ def groups(self):
     def on_Assign(self, stmt):
         self.unify(*stmt._lhs_signals())
 
+    def on_Assert(self, stmt):
+        self.unify(*stmt._lhs_signals())
```
whitequark: You can probably just do `on_Assert = on_Assign`.

nmigen/hdl/xfrm.py
```diff
@@ -352,6 +384,16 @@ def on_Assign(self, stmt):
         if any_lhs_signal in self.signals:
             return stmt
 
+    def on_Assert(self, stmt):
```
whitequark: Same here.

nmigen/hdl/ast.py
```diff
@@ -826,6 +827,65 @@ def __repr__(self):
...
+    file, line = src_loc
+    return "{}:{}".format(file, line)
+
+
+class Assume(Statement):
+    def __init__(self, test, _en=None, _check=None):
+        self.test = Value.wrap(test)
+        if _en is None:
+            self._en = Signal(1, reset_less=True,
+                              name="assume${}$en".format(src(self.test.src_loc)))
```
whitequark: This produces utterly *gigantic* IL statements like `connect \A \assert$/home/whitequark/Projects/Boneless-CPU/boneless/gateware/formal.py:25$check. I think it's sufficient to call them something like `$assume$en`

whitequark:
Merged.

PR #19, opened 2019-01-01
Fix #17 (Add a more user-friendly error message when the yosys binary can't be found)

### LINKED ISSUE #17: Add a more user-friendly error message when the yosys binary can't be found
When the `YOSYS` environment variable isn't defined, the fallback `yosys` is used as the name of the yosys executable ([src](https://github.com/m-labs/nmigen/blob/master/nmigen/back/verilog.py#L16)). When this fails (i.e. the environment variable is not defined and `yosys` is not in the user's `PATH`), a file not found error is thrown.

Suggestion: add a more user-friendly error message that prompts the user to define/check the `YOSYS` environment variable or add yosys to their `PATH`

The6P4C:
Should fix #17.

Not sure if this is the error message format you'd like, couldn't find any examples of raising a similar error. Keeps the original `FileNotFound` error intact by chaining the `RuntimeError` to it.

Tested on both Linux and Windows.

### DIFF
--- a/nmigen/back/verilog.py
+++ b/nmigen/back/verilog.py
@@ -13,12 +13,14 @@ class YosysError(Exception):
 
 def convert(*args, **kwargs):
     il_text = rtlil.convert(*args, **kwargs)
-    popen = subprocess.Popen([os.getenv("YOSYS", "yosys"), "-q", "-"],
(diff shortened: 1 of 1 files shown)


whitequark:
Added an even better error in 06faeee.

PR #21, opened 2019-01-06
Give the top level scope a name to fix vcd file production

adamgreig:
At present the `root_fragment` in the pysim `hierarchy` has an empty `scope=()`, and when this is added to the vcd file we get:
```vcd
$scope module  $end
```
i.e. the top-level module has no name. This breaks gtkwave somewhat as you can't re-select the top level after selecting a submodule:

image

This patch sets the `root_fragment`'s scope to `"top"`, producing a vcd with a useful hierarchy:

image

I can't see an obvious way to get a more useful name than "top" for the root fragment, but if one exists that would be even better.

### DIFF
--- a/nmigen/back/pysim.py
+++ b/nmigen/back/pysim.py
@@ -448,7 +448,7 @@ class Simulator:
                     add_fragment(subfragment, (*scope, "#{}".format(index)))
                 else:
                     add_fragment(subfragment, (*scope, name))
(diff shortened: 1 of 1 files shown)


whitequark:
> This breaks gtkwave somewhat as you can't re-select the top level after selecting a submodule:

Actually, you can. Click on e.g. `rdport` again after selecting it. But I guess your solution is better.

PR #25, opened 2019-01-15
Support for yosys' $anyconst and $anyseq cells

cr1901:
`AnyConst` and `AnySeq` correspond to to the `$anyconst` and `$anyseq` RTLIL cells. They are useful for formal verification to represent "any value that doesn't change", and "any value that _can_ change to another arbitrary value each clock cycle" respectively.

Usage is analogous to declaring a `Constant` or `Signal`:

```
self.foo = AnyConst(2)
self.bar = AnySeq(16)
```

### DIFF
--- a/nmigen/hdl/ast.py
+++ b/nmigen/hdl/ast.py
@@ -9,7 +9,7 @@ from ..tools import *
 
 
 __all__ = [
-    "Value", "Const", "C", "Operator", "Mux", "Part", "Slice", "Cat", "Repl",
+    "Value", "Const", "C", "AnyConst", "AnySeq", "Operator", "Mux", "Part", "Slice", "Cat", "Repl",
     "Array", "ArrayProxy",
     "Signal", "ClockSignal", "ResetSignal",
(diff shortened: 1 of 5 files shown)


nmigen/hdl/ast.py
```diff
@@ -254,6 +254,38 @@ def __repr__(self):
...
+
+    def shape(self):
+        return self.nbits, self.signed
+
+    def _rhs_signals(self):
+        return ValueSet()
+
+
+class AnyConst(AnyValue):
+    def __init__(self, shape):
```
whitequark: This is a no-op--this class does not need the `__init__` method. (Though, once you remove it, you should change `src_loc_at` above.)

nmigen/hdl/ast.py
```diff
@@ -254,6 +254,38 @@ def __repr__(self):
...
+        super().__init__(shape)
+
+    def __repr__(self):
+        return "(anyconst {}'{})".format(self.nbits, "s" if self.signed else "")
+
+
+class AnySeq(AnyValue):
+    def __init__(self, shape):
+        super().__init__(shape)
```
whitequark: Same here.

PR #36, opened 2019-02-21
setup.py: constrain Python version

peteut:
Installation should be constraint to supported Python versions, using `python_requires`, refer to [1] for details.

[1] https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,7 @@ setup(
     license="BSD",
     install_requires=["pyvcd>=0.1.4", "bitarray"],
     packages=find_packages(),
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -16,6 +16,7 @@
     license="BSD",
     install_requires=["pyvcd>=0.1.4", "bitarray"],
     packages=find_packages(),
+    python_requires=">3.6",
```
whitequark: `>=3.6`?

Issue #29, opened 2019-01-27
ResetInserter &c cannot be used on Modules


whitequark:
For added fun, they cannot be used on *CompatModule*s either.

PR #38, opened 2019-03-03
Fix CDC issue in async FIFO: ensure that Gray code pointers are registered before resynchronisation.

Wren6991:
The AsyncFIFO class is subtly different from the Clifford Cummings paper referenced in the source code. Here is a diagram showing the difference:

In the current nMigen code, there is a path that goes from:

- `produce` counter register in the write domain
- -> `GrayEncoder` 
- -> `MultiReg` in the read domain.

The problem is that there is a narrow window after a write clock edge where all of the Gray pointer bits may toggle, due to static hazards in the encoder, since all encoder _inputs_ can toggle simultaneously. If you sample this with an asynchronous clock, it's possible to see a completely wrong value for the `produce` pointer, which can cause irreversible corruption/loss of data.

I've only shown this for the `produce` pointer, but the same is true in the other direction too. The fix is to drive the Gray encoder from the binary counter register's input, rather than its output, and register the encoder output before resynchronising. From a synchronous point of view this behaves the same as the old Gray count, so the full/empty logic stays the same.

This bug will only be observed with a low probability, and may not be observed at all, e.g. if your two asynchronous clocks share some fixed rational relationship with a common clock source through PLLs, which affects the distribution of the relative phase of one observed at the rising edge of the other. However it _is_ a legitimate bug, with fatal consequences.

This code is not ready to merge:

- So far I've only run the FIFO smoke test on it. I haven't yet figured out how to run the formal test :( and it also needs soak testing on an FPGA board with two _completely_ separate oscillators.
- The pointer `MultiReg`s should ideally be resettable (although that is a different conversation!)
- I added a helper class, `GrayBinCounter`, which is currently only used by `AsyncFIFO`. It contains some extra detail, like putting a `no_retiming` attribute on the Gray register.
    - Currently I'm passing the domain name into the `__init__` of this class, like `MultiReg`, but I saw references elsewhere to the `DomainRenamer` class, which seems cleaner. However it seems you can only call this on `Fragment`s, not `Modules`, and I currently don't know what a `Fragment` is :)
    - If you would prefer not to have an extra helper class, I can give an alternative patch that just modifies `AsyncFIFO`. It just looks a bit copy-pastey.

I'm really new to nMigen, so this code definitely needs a close review.

### DIFF
--- a/nmigen/lib/fifo.py
+++ b/nmigen/lib/fifo.py
@@ -264,6 +264,56 @@ class SyncFIFOBuffered(FIFOInterface):
 
         return m
 
+# A helper module for AsyncFIFO
+class GrayBinCounter:
+    __doc__ = """Synchronous counter with both binary and Gray code output.
+    Both outputs count synchronously while ``incr`` is asserted.
+
+    ``gray`` being registered is necessary for clock domain crossing.
+    Putting a GrayEncoder on the output of a simple binary counter creates
+    static hazards on (almost) all output bits; Gray code's CDC safety
+    is based on a guarantee that only a single bit transitions at once.
+
+    Parameters
+    ----------
+    width : int
+        Bit width of both outputs
+    domain : str
+        Clock domain which count is synchronous to. Defaults to "sync".
+    Attributes
+    ----------
+    incr : Signal(1)
+        The counter increments when ``incr`` is high.
+    bin : Signal(width)
+        Binary count value
+    gray: Signal(width)
+        Gray-coded count value
+    """
+
+    def __init__(self, width, domain="sync"):
+        self.width = width
+        self.domain = domain
+
+        self.incr = Signal()
+        self.bin = Signal(width)
+        self.gray = Signal(width, attrs={"no_retiming": True})
+
+    def elaborate(self, platform):
+        m = Module()
+
+        enc = m.submodules.enc = GrayEncoder(self.width)
+        bin_next = self.bin + self.incr
+        gray_next = Signal(self.width)
+
+        m.d.comb += [
+            enc.i.eq(bin_next),
+            gray_next.eq(enc.o)
+        ]
+        m.d[self.domain] += [
+            self.bin.eq(bin_next),
+            self.gray.eq(gray_next)
+        ]
+        return m
 
 class AsyncFIFO(FIFOInterface):
     __doc__ = FIFOInterface._doc_template.format(
@@ -297,28 +347,33 @@ class AsyncFIFO(FIFOInterface):
 
         m = Module()
 
+        do_write = self.writable & self.we
+        do_read  = self.readable & self.re
+
         produce_w_bin = Signal(self._ctr_bits)
         produce_w_gry = Signal(self._ctr_bits)
         produce_r_gry = Signal(self._ctr_bits)
-        produce_enc = m.submodules.produce_enc = \
-            GrayEncoder(self._ctr_bits)
         produce_cdc = m.submodules.produce_cdc = \
             MultiReg(produce_w_gry, produce_r_gry, odomain="read")
+        produce_ctr = m.submodules.produce_ctr = \
+            GrayBinCounter(self._ctr_bits, domain="write")
         m.d.comb += [
-            produce_enc.i.eq(produce_w_bin),
-            produce_w_gry.eq(produce_enc.o),
+            produce_ctr.incr.eq(do_write),
+            produce_w_bin.eq(produce_ctr.bin),
+            produce_w_gry.eq(produce_ctr.gray)
         ]
 
         consume_r_bin = Signal(self._ctr_bits)
         consume_r_gry = Signal(self._ctr_bits)
         consume_w_gry = Signal(self._ctr_bits)
-        consume_enc = m.submodules.consume_enc = \
-            GrayEncoder(self._ctr_bits)
         consume_cdc = m.submodules.consume_cdc = \
             MultiReg(consume_r_gry, consume_w_gry, odomain="write")
+        consume_ctr = m.submodules.consume_ctr = \
+            GrayBinCounter(self._ctr_bits, domain="read")
         m.d.comb += [
-            consume_enc.i.eq(consume_r_bin),
-            consume_r_gry.eq(consume_enc.o),
+            consume_ctr.incr.eq(do_read),
+            consume_r_bin.eq(consume_ctr.bin),
+            consume_r_gry.eq(consume_ctr.gray),
         ]
 
         m.d.comb += [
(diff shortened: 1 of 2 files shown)


Wren6991:
Yep looks like I've broken something :D Will update once I've got formal running on my own machine, and fixed the issue.

Wren6991:
image

The reset values look nonsensical to me, e.g. 010 (bin) != 010 (gray). Maybe the fact it passed before was a lucky break due to the bin and gray pointers being driven by the same register, with the same initial value. I need to dig into this formal stuff more.

Wren6991:
I've added an additional commit which asserts FIFO resets whilst `$initstate` is true, in FIFOContractSpec. On my machine I can now run `coverage run -m unittest discover` and all tests pass.

Since this fix is outside the scope of this PR, I am happy to open a separate one if you would prefer.

whitequark:
First, thanks for finding the bug! This is indeed very important, and I'm not sure how I missed it. It is definitely not present in Migen, it's an nMigen bug.

> I've added an additional commit which asserts FIFO resets whilst `$initstate` is true, in FIFOContractSpec. On my machine I can now run `coverage run -m unittest discover` and all tests pass.

Unfortunately you've significantly weakened the contract by adding a reset. Before your changes, the FIFO contract would test an *arbitrary* initial state (any block RAM contents and any pointers). After your changes the initial state is always defined, so e.g. a FIFO that would pass just two elements through it and then hang would pass it.

whitequark:
Can you check if the following diff also fixes the bug, please?
```diff
diff --git a/nmigen/lib/fifo.py b/nmigen/lib/fifo.py
index 3af5c51..d4f16e1 100644
--- a/nmigen/lib/fifo.py
+++ b/nmigen/lib/fifo.py
@@ -304,10 +304,8 @@ class AsyncFIFO(FIFOInterface):
             GrayEncoder(self._ctr_bits)
         produce_cdc = m.submodules.produce_cdc = \
             MultiReg(produce_w_gry, produce_r_gry, odomain="read")
-        m.d.comb += [
-            produce_enc.i.eq(produce_w_bin),
-            produce_w_gry.eq(produce_enc.o),
-        ]
+        m.d.comb  += produce_enc.i.eq(produce_w_bin)
+        m.d.write += produce_w_gry.eq(produce_enc.o)
 
         consume_r_bin = Signal(self._ctr_bits)
         consume_r_gry = Signal(self._ctr_bits)
@@ -316,17 +314,15 @@ class AsyncFIFO(FIFOInterface):
             GrayEncoder(self._ctr_bits)
         consume_cdc = m.submodules.consume_cdc = \
             MultiReg(consume_r_gry, consume_w_gry, odomain="write")
-        m.d.comb += [
-            consume_enc.i.eq(consume_r_bin),
-            consume_r_gry.eq(consume_enc.o),
-        ]
+        m.d.comb  += consume_enc.i.eq(consume_r_bin)
+        m.d.read  += consume_r_gry.eq(consume_enc.o)
 
         m.d.comb += [
             self.writable.eq(
-                (produce_w_gry[-1]  == consume_w_gry[-1]) |
-                (produce_w_gry[-2]  == consume_w_gry[-2]) |
-                (produce_w_gry[:-2] != consume_w_gry[:-2])),
-            self.readable.eq(consume_r_gry != produce_r_gry)
+                (produce_enc.o[-1]  == consume_w_gry[-1]) |
+                (produce_enc.o[-2]  == consume_w_gry[-2]) |
+                (produce_enc.o[:-2] != consume_w_gry[:-2])),
+            self.readable.eq(consume_enc.o != produce_r_gry)
         ]
 
         do_write = self.writable & self.we
diff --git a/nmigen/test/test_lib_fifo.py b/nmigen/test/test_lib_fifo.py
index c7f4c28..e70d7be 100644
--- a/nmigen/test/test_lib_fifo.py
+++ b/nmigen/test/test_lib_fifo.py
@@ -262,8 +262,8 @@ class FIFOFormalCase(FHDLTestCase):
         # self.assertFormal(FIFOModelEquivalenceSpec(fifo, rdomain="read", wdomain="write"),
         #                   mode="bmc", depth=fifo.depth * 3 + 1)
         self.assertFormal(FIFOContractSpec(fifo, rdomain="read", wdomain="write",
-                                           bound=fifo.depth * 4 + 1),
-                          mode="hybrid", depth=fifo.depth * 4 + 1)
+                                           bound=(fifo.depth + 1) * 4),
+                          mode="hybrid", depth=(fifo.depth + 1) * 4)
 
     def test_async(self):
         self.check_async_fifo(AsyncFIFO(width=8, depth=4))
```

Wren6991:
> I'm not sure how I missed it
Possibly distracted by CC's name!

> Unfortunately you've significantly weakened the contract by adding a reset. Before your changes, the FIFO contract would test an _arbitrary_ initial state (any block RAM contents and any pointers). After your changes the initial state is always defined, so e.g. a FIFO that would pass just two elements through it and then hang would pass it.

Ah okay, I see the problem.

Wouldn't you normally test this with a separate induction pass? The failure I saw was due to the design starting in an unreachable state. You are talking about reachable states, which should be testable by induction after a sufficiently long BMC. I haven't used SymbiYosys before, but have used yosys-smtbmc with a different solver backend, and always performed BMC followed by induction.

whitequark:
> Wouldn't you normally test this with a separate induction pass?

Yes, but I never got induction to work on all the FIFO designs. This oddball BMC is the best I could do.

Wren6991:
That diff fixes the bug, but does introduce an extra cycle of latency in each direction. Immediate problem is defused :)

> Yes, but I never got induction to work on all the FIFO designs. This oddball BMC is the best I could do.

Sounds like your experience with formal has been similar to mine... if this would be a valuable thing then I can have a play with it. It would be nice at some point in the future to win back that additional latency cycle, without breaking the test!

Thank you for taking a look at my patch.

whitequark:
> That diff fixes the bug, but does introduce an extra cycle of latency in each direction. Immediate problem is defused :)

Oh, I see! I will do something about it. Thanks for pointing out!

> * Currently I'm passing the domain name into the `__init__` of this class, like `MultiReg`, but I saw references elsewhere to the `DomainRenamer` class, which seems cleaner. However it seems you can only call this on `Fragment`s, not `Modules`, and I currently don't know what a `Fragment` is :)

This is an open issue, #29. For now I will do the copy-pastey thing.

Wren6991:
> This is an open issue, #29. For now I will do the copy-pastey thing.

Ah okay. I had a quick look at that area of code (Module etc) and decided I needed to learn some GoF patterns first!

If it's any help at all, this is what my "flat" patch looked like: [link to commit](https://github.com/Wren6991/nmigen/commit/78b6cd032a5aaa22b3fc3823cfec6e633947493f)

But this had the same issue with the FIFO contract test.

whitequark:
> I had a quick look at that area of code (Module etc) and decided I needed to learn some GoF patterns first!

I'm not really a huge fan of GoF. That code doesn't need patterns so much as one horrible hack, and I can't decide exactly how horrible it's going to be. So it's not done just yet.

whitequark:
Fixed by 40273178352d9, without any added latency. I've solved this properly by adding appropriate `Assume`s. You are right that this pattern should be extracted into a `GrayCounter`; it was in Migen, but I did not understand exactly why.

Issue #28, opened 2019-01-26
Multiclock simulation broken


whitequark:
Repro:
```python
from nmigen import *
from nmigen.back.pysim import *


m = Module()
m.domains += ClockDomain("input")
m.domains += ClockDomain("output")

with Simulator(m, vcd_file=open("test.vcd", "w")) as sim:
    sim.add_clock(3e-6, domain="input")
    sim.add_clock(5e-6, domain="output")
    sim.run_until(1, run_passive=True)
```

RobertBaruch:
Apparently when the two clocks need to change at the same time (here, at 7.5us), at timestamp 7.5, the scheduled events are at 7.5 and 10, with a delta of 1e-10. Since 7.5-min(7.5, 10)=0 <= 1e-10, a `DeadlineError` is raised.

You probably already know this :)  

```
$ python3 simclk.py
step
  curr_dirty any
    wait_deadline {}, ts 0.0, delta 0.0
step
  processes > suspended
step
  processes > suspended
step
  run passive
step
  curr_dirty any
    wait_deadline {<generator object Simulator.add_clock.<locals>.clk_process at 0x7f447bd295c8>: 2.5, <generator object Simulator.add_clock.<locals>.clk_process at 0x7f447bd29620>: 3.0}, ts 1.5, delta 1e-10
     wait_deadline.values dict_values([2.5, 3.0])
step
  run passive
step
  curr_dirty any
    wait_deadline {<generator object Simulator.add_clock.<locals>.clk_process at 0x7f447bd29620>: 3.0, <generator object Simulator.add_clock.<locals>.clk_process at 0x7f447bd295c8>: 5.0}, ts 2.5, delta 1e-10
     wait_deadline.values dict_values([3.0, 5.0])
step
  run passive
step
  curr_dirty any
    wait_deadline {<generator object Simulator.add_clock.<locals>.clk_process at 0x7f447bd295c8>: 5.0, <generator object Simulator.add_clock.<locals>.clk_process at 0x7f447bd29620>: 4.5}, ts 3.0, delta 0.0
     wait_deadline.values dict_values([5.0, 4.5])
step
  run passive
step
  curr_dirty any
    wait_deadline {<generator object Simulator.add_clock.<locals>.clk_process at 0x7f447bd295c8>: 5.0, <generator object Simulator.add_clock.<locals>.clk_process at 0x7f447bd29620>: 6.0}, ts 4.5, delta 1e-10
[...]

PR #40, opened 2019-03-04
WIP: Expand and document lib.cdc

Wren6991:
Very much a work in progress, but I wanted to open early to get feedback on the approach/direction. I'm porting/rewriting some of the missing modules from migen.genlib.cdc.

Currently in this patch set:

- Add docstrings to the existing modules. I tried to copy the "house style" I saw elsewhere.
- Basic checking to give more meaningful traces when you do something like request a MultiReg of length 0. Again I tried to copy the idioms I saw elsewhere
- Implement PulseSynchronizer
- Implement Gearbox

Known issues:
- Smoke tests only; no formal checks
- Checks are limited by single-clock simulation -- maybe it was a bad idea to play with the CDC library first :)
- I'm passing in the domain names rather than using DomainRenamer. Not sure whether it's best to leave this PR open until this is in place, or merge early so people can start hacking on this. Not my decision :)
- The storage size calculation in this version of Gearbox is different to the one in Migen. IMO the old one was unsafe, but this one may be too conservative. I could also be plain wrong!
- (edit:) The output mux on the Gearbox is driven from two clock domains. This is a legitimate thing to do here, but I saw mention somewhere that doing this should cause an error.

I definitely still intend to port:
- BusSynchronizer
- ElasticBuffer

And we probably ought to do something about GrayCounter at some point, but I think it's less important than the others.

As a general style question, there seem to be two ways of doing module wiring in `nmigen.lib` at the moment. One is to pass external signals into `__init__`, which gives you more of a Verilog-style instantiation. The other is to create "port" signals in `__init__`, which the parent then wires up during elaboration. I've used the second style because it seems a bit cleaner, and doesn't require creating extraneous signals like in Verilog, but not sure if this is the right thing to do?

### DIFF
--- a/nmigen/lib/cdc.py
+++ b/nmigen/lib/cdc.py
@@ -7,6 +7,7 @@ __all__ = [
     "ResetSynchronizer",
     "PulseSynchronizer",
     "BusSynchronizer",
+    "ElasticBuffer",
     "Gearbox"
 ]
 
@@ -260,6 +261,66 @@ class BusSynchronizer(Elaboratable):
             m.d[self.odomain] += self.o.eq(buf_o)
 
         return m
+
+class ElasticBuffer(Elaboratable):
+    """Pass data between two clock domains with the same frequency, and bounded phase difference.
+
+    Increasing the storage depth increases tolerance for clock wander and jitter, but still within
+    some bound. For less-well-behaved clocks, consider AsyncFIFO.
+
+    Parameters
+    ----------
+    width : int > 0
+        Width of databus to be resynchronized
+    depth : int > 1
+        Number of storage elements in buffer
+    idomain : str
+        Name of input clock domain
+    odomain : str
+        Name of output clock domain
+
+    Attributes
+    ----------
+    i : Signal(width)
+        Input data bus
+    o : Signal(width)
+        Output data bus
+    """
+    def __init__(self, width, depth, idomain, odomain):
+        if not isinstance(width, int) or width < 1:
+            raise TypeError("width must be a positive integer, not '{!r}'".format(width))
+        if not isinstance(depth, int) or depth <= 1:
+            raise TypeError("depth must be an integer > 1, not '{!r}'".format(depth))
+
+        self.i = Signal(width)
+        self.o = Signal(width)
+        self.width = width
+        self.depth = depth
+        self.idomain = idomain
+        self.odomain = odomain
+
+    def elaborate(self, platform):
+        m = Module()
+
+        wptr = Signal(max=self.depth, reset=self.depth // 2)
+        rptr = Signal(max=self.depth)
+        m.d[self.idomain] += wptr.eq(_incr(wptr, self.depth))
+        m.d[self.odomain] += rptr.eq(_incr(rptr, self.depth))
+
+        storage = Memory(self.width, self.depth)
+        wport = m.submodules.wport = storage.write_port(domain=self.idomain)
+        rport = m.submodules.rport = storage.read_port(domain=self.odomain)
+
+        m.d.comb += [
+            wport.en.eq(1),
+            wport.addr.eq(wptr),
+            wport.data.eq(self.i),
+            rport.addr.eq(rptr),
+            self.o.eq(rport.data)
+        ]
+
+        return m
+
 
 class Gearbox(Elaboratable):
     """Adapt the width of a continous datastream.
--- a/nmigen/lib/cdc.py
+++ b/nmigen/lib/cdc.py
@@ -2,7 +2,13 @@ from .. import *
 from math import gcd
 
 
-__all__ = ["MultiReg", "ResetSynchronizer", "PulseSynchronizer", "Gearbox"]
+__all__ = [
+    "MultiReg",
+    "ResetSynchronizer",
+    "PulseSynchronizer",
+    "BusSynchronizer",
+    "Gearbox"
+]
 
 
 def _incr(signal, modulo):
@@ -135,7 +141,6 @@ class PulseSynchronizer(Elaboratable):
 
     Parameters
     ----------
-
     idomain : str
         Name of input clock domain.
     odomain : str
@@ -169,6 +174,92 @@ class PulseSynchronizer(Elaboratable):
 
         return m
 
+class BusSynchronizer(Elaboratable):
+    """Pass a multi-bit signal safely between clock domains.
+
+    Ensures that all bits presented at ``o`` form a single word that was present synchronously at
+    ``i`` in the input clock domain (unlike direct use of MultiReg).
+
+    Parameters
+    ----------
+    width : int > 0
+        Width of the bus to be synchronized
+    idomain : str
+        Name of input clock domain
+    odomain : str
+        Name of output clock domain
+    sync_stages : int >= 2
+        Number of synchronisation stages used in the req/ack pulse synchronizers. Lower than 2 is
+        unsafe. Higher values increase safety for high-frequency designs, but increase latency too.
+    timeout : int >= 0
+        The request from idomain is re-sent if ``timeout`` cycles elapse without a response.
+        ``timeout`` = 0 disables this feature.
+
+    Attributes
+    ----------
+    i : Signal(width), in
+        Input signal, sourced from ``idomain``
+    o : Signal(width), out
+        Resynchronized version of ``i``, driven to ``odomain``
+    """
+    def __init__(self, width, idomain, odomain, sync_stages=2, timeout = 127):
+        if not isinstance(width, int) or width < 1:
+            raise TypeError("width must be a positive integer, not '{!r}'".format(width))
+        if not isinstance(sync_stages, int) or sync_stages < 2:
+            raise TypeError("sync_stages must be an integer > 1, not '{!r}'".format(sync_stages))
+        if not isinstance(timeout, int) or timeout < 0:
+            raise TypeError("timeout must be a non-negative integer, not '{!r}'".format(timeout))
+
+        self.i = Signal(width)
+        self.o = Signal(width, attrs={"no_retiming": True})
+        self.width = width
+        self.idomain = idomain
+        self.odomain = odomain
+        self.sync_stages = sync_stages
+        self.timeout = timeout
+
+    def elaborate(self, platform):
+        m = Module()
+        if self.width == 1:
+            m.submodules += MultiReg(self.i, self.o, odomain=self.odomain, n=self.sync_stages)
+            return m
+
+        req = Signal()
+        ack_o = Signal()
+        ack_i = Signal()
+
+        sync_io = m.submodules.sync_io = \
+            PulseSynchronizer(self.idomain, self.odomain, self.sync_stages)
+        sync_oi = m.submodules.sync_oi = \
+            PulseSynchronizer(self.odomain, self.idomain, self.sync_stages)
+
+        if self.timeout != 0:
+            countdown = Signal(max=self.timeout, reset=self.timeout)
+            with m.If(ack_i | req):
+                m.d[self.idomain] += countdown.eq(self.timeout)
+            with m.Else():
+                m.d[self.idomain] += countdown.eq(countdown - countdown.bool())
+
+        start = Signal(reset=1)
+        m.d[self.idomain] += start.eq(0)
+        m.d.comb += [
+            req.eq(start | ack_i | (self.timeout != 0 and countdown == 0)),
+            sync_io.i.eq(req),
+            ack_o.eq(sync_io.o),
+            sync_oi.i.eq(ack_o),
+            ack_i.eq(sync_oi.o)
+        ]
+
+        buf_i = Signal(self.width, attrs={"no_retiming": True})
+        buf_o = Signal(self.width)
+        with m.If(ack_i):
+            m.d[self.idomain] += buf_i.eq(self.i)
+        sync_data = m.submodules.sync_data = \
+            MultiReg(buf_i, buf_o, odomain=self.odomain, n=self.sync_stages - 1)
+        with m.If(ack_o):
+            m.d[self.odomain] += self.o.eq(buf_o)
+
+        return m
 
 class Gearbox(Elaboratable):
     """Adapt the width of a continous datastream.
--- a/nmigen/lib/cdc.py
+++ b/nmigen/lib/cdc.py
@@ -31,15 +31,15 @@ class MultiReg(Elaboratable):
         Signal to be resynchronised
     o : Signal(), out
         Signal connected to synchroniser output
-    odomain : str
-        Name of output clock domain
+    cd_o : str
+        Name of output (capturing) clock domain
     n : int
         Number of flops between input and output.
     reset : int
         Reset value of the flip-flops. On FPGAs, even if ``reset_less`` is True, the MultiReg is
         still set to this value during initialization.
     reset_less : bool
-        If True (the default), this MultiReg is unaffected by ``odomain`` reset.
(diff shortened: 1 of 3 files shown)


Wren6991:
Last commit is just to fix a small coverage blackspot in the existing code. I was having far too much fun with the graphs on Codecov

Wren6991:
BusSynchronizer is as per Migen except:
- number of synchronisation stages is optionally parameterised
- timeout retry is optional by setting timeout to 0 (since it is a bit risky, can cause metastabilities in capturing domain)
- Timer is not factored out into a module. nMigen does not have a `misc` library right now

Again it needs formal checks, and tests are devalued by running them in one clock domain.

Wren6991:
Add ElasticBuffer. Faithful reproduction of the Migen version, except without the reset synchronisers (since these would normally be per-domain, and there is still some code motion wrt reset/clock domains in nMigen as a whole).

Force push due to finding small issues in earlier commits, e.g. gearbox increment didn't work for non-pow-2 chunk counts (which can't be tested in one clock domain, I just spotted it while reading the code). I've tried to keep it so that each commit can be separately checked out, tested and reviewed. If you'd rather I just push fixes, let me know :)

Also, I've copied the `_incr` used in `lib.fifo`. Might be worth factoring this out, but that's outside of this PR.

whitequark:
> * I'm passing in the domain names rather than using DomainRenamer. Not sure whether it's best to leave this PR open until this is in place, or merge early so people can start hacking on this. Not my decision :)

I'll try to fix DomainRenamer to work on Modules as soon as possible!

> As a general style question, there seem to be two ways of doing module wiring in `nmigen.lib` at the moment. One is to pass external signals into `__init__`, which gives you more of a Verilog-style instantiation. The other is to create "port" signals in `__init__`, which the parent then wires up during elaboration. I've used the second style because it seems a bit cleaner, and doesn't require creating extraneous signals like in Verilog, but not sure if this is the right thing to do?

I use the second style, with some exceptions like MultiReg. The reason is that the signals aren't always completely independent, for example you might want to pass a width and have a set of signals with that width, double that width, etc created for you. Also, constructors with a large number of arguments are unwieldy and bug-prone, especially if you change the constructor later.

Wren6991:
> I'll try to fix DomainRenamer to work on Modules as soon as possible!

Ok, no sweat, I'll rebase + fix as and when :)

Wren6991:
Rebased on latest master and fixed up docstrings:

- Reflowed to 100 characters
- Fixed ReST formatting
- Americanized spellings to be consistent with class names
- Some edits for clarity

I also stared at the code for a while, only one minor fix which was something that really oughtn't be retimed, but didn't have the `no_retiming` attribute. Not saying there's nothing to fix, just that it's reached the point where squinting at my own code is nonproductive.

Domains are passed as strings, will fix this after the `xfrm.py` changes go in.

whitequark:
@Wren6991 Done!

Wren6991:
Thanks :) I've fixed up PulseSynchronizer and Gearbox so far. Unfortunately I'm away on a trip at the moment and don't have all the necessary tools set up on my work laptop, so it'll take me a while to get the fixes tested and pushed.

For consistency I'm using "read" and "write" for default domain names (matching AsyncFIFO), but there is a not-totally-frivolous argument to use something like "i" and "o" instead:

- Matches the signal naming convention used in e.g. `lib/coding.py`. Makes sense for `i` to be in `"i"`, `o` in `"o"` etc
- Having domains be the same number of characters is super nice. I have bad habits from RTL of lining up blocks of assignments in my code, and it's nice if this happens without inserting whitespace
- `DomainRenamer({"i": "gpu", "o": "display")(MyCDCModule())` is just a bit pithier

Was just something that occurred to me when editing the files, it's not up to me.

whitequark:
> Thanks :) I've fixed up PulseSynchronizer and Gearbox so far. Unfortunately I'm away on a trip at the moment and don't have all the necessary tools set up on my work laptop, so it'll take me a while to get the fixes tested and pushed.
> 
> For consistency I'm using "read" and "write" for default domain names (matching AsyncFIFO), but there is a not-totally-frivolous argument to use something like "i" and "o" instead:

Now that I think more about it, I think it doesn't make sense to use `DomainRenamer` for *synchronization primitives*. The reason being, for these primitives, you always want to rename them, so requiring *each caller* to do that work is clearly superfluous, and it results in unnecessarily hard to read code too.

There's a good argument that `PulseSynchronizer` should work the same as `AsyncFIFO`. I agree. I think `AsyncFIFO` should use explicit domain specification as well! Pretty much every instance of `AsyncFIFO` is immediately followed by a `DomainRenamer` call.

WDYT?

Wren6991:
So a flow could be:

- If you are writing a block of synchronous logic, you use `sync`, and then apply `DomainRenamer` to yeet it into the correct domain
- Knowledge of domains is confined to a specific region of your code, which places synchronous blocks into the correct domains and also glues them together with CDC modules
- Since CDC + `DomainRenamer` is actually a single constructor in a trenchcoat, skip the extra step and pass domains in directly

If so, that feels cleaner, yeah!

whitequark:
Yep!

> to yeet it into the correct domain

Incredible.

Wren6991:
Awesome. I guess the next step for this PR is a lot more testing then. I did note #28, which I guess I could start to take a look at at some point, as this bumps against it. It would be good to learn about that part. Formal would be good too.

For CDC it would also be cool to do some soak testing on real FPGAs, to catch genuine CDC bugs rather than just state machine bugs (albeit they're difficult to debug once found). Not sure if this is something you already had ideas on.

whitequark:
> For CDC it would also be cool to do some soak testing on real FPGAs

The CDC primitives in oMigen are extremely well tested. I think we mainly need to make sure the nMigen ones don't stray from them, like it happened with AsyncFIFO...

Wren6991:
Ok that's a good point. Will read through the oMigen library again later today. For the most part it's pretty faithful but mistakes creep in!

There are things in the old library that seemed odd though, e.g. for Gearbox with `lcm(iwidth, owidth) == iwidth`, the write port will have depth 2. The pointers start 180° apart, but there is no guard period between final o read on an ichunk and the next write to that ichunk, or conversely, first read from an ichunk and the preceding write. You're at the mercy of dice roll on the two reset synchronisers. I'm currently using a different storage size calculation because of this (it can end up larger or smaller than oMigen depending on ratio).

I'll definitely make sure the logic is all ok, and then maybe raise questions about things like the above.

whitequark:
Sounds good!

Wren6991:
I did a side-by-side of oMigen vs current state of my code

- `PulseSynchronizer`
    - Actual logic identical
    - Number of sync stages in `MultiReg` is parameterised.
    - Currently I don't have the `reset_less` attribute on the non-CDC flops in this module, but the MultiReg defaults to `reset_less=True`. It would probably be better to parameterise this, and either have everything or nothing `reset_less`
- `BusSynchronizer`
    - Handshake logic is identical: req/ack with two PulseSynchronizers
    - Number of sync stages parameterised
    - Option to disable timeout, as it has small chance of temporarily breaking the handshake for oclock << iclock (or oclock sometimes gated); multiple `req`s can be in flight!
    - **I removed one layer of flops from data CDC path. See diagram**
    - Again some confusion over what should/should not be `reset_less`. Currently all synchronous logic is resettable but the `MultiReg`s are not
    - Found an issue with my code: `ibuffer` clock enable was driven by only `ack_o` in oMigen, but mine was more permissive, and also enabled on timer restarts and start-of-day flag. Think this is ok, but I've fixed it to be exactly like oMigen.
- `ElasticBuffer`
    - Logic is faithful reproduction of oMigen
    - oMigen code also has some reset structure inside. It ORs together the reset from each domain, and resynchronises this new reset internally. I think this is problematic because reset *assertion* is asynchronous, so e.g. an input-domain reset can inject metastabilities into the output domain. This is better handled by proper reset sequencing at system level. WDYT?
- `GearBox`
    - Same comment on resets as `ElasticBuffer`
    - Actual logic is faithful to oMigen
    - Storage size calculation is different in my version: oMigen takes LCM of two widths, and doubles it unconditionally. This is unnecessarily large for many coprime widths (e.g. 5:4 for video) and for `lcm(i, o) in (i, o)` it seems too small to me. My code starts with `lcm(i, o)` and doubles it until the depth of each port is at least four, logic being that this gives you one chunk of no-man's-land on each side between the chunk you are reading/writing and the chunk the other side is writing/reading. WDYT?

**BusSynchronizer datapath**

I drew a rough box diagram of oMigen `BusSynchronizer`:

image

My problem with this is that there is a 3-flop delay for both the request and the data. (1 flop in `idomain`, two in `odomain`). This means there is a race between the `obuffer` clock enable and the data. One reason I think this has been safe so far is that there is a 1-LUT delay on the `obuffer` clock enable due to the XOR gate detecting the transitions. This is probably just enough to ensure that enable is not observed before data valid.

I've fixed the race by removing one stage from the datapath `MultiReg`. Note there are still two non-retimable layers of flops in `odomain` on the datapath, and also, `sync_stages` can be increased to stretch out the entire `BusSynchronizer`.

whitequark:
@Wren6991 Sorry, I completely missed your update, GitHub decided to not send me an email for it for some reason.

nmigen/lib/cdc.py
```diff
@@ -70,7 +72,27 @@ def elaborate(self, platform):
...
+    Output `assertion` is asynchronous, so the local clock need not be free-running.
+
+    Parameters
+    ----------
+    arst : Signal(1), out
+        Asynchronous reset signal, to be synchronized.
+    domain : str
+        Name of domain to synchronize reset to.
+    n : int, >=1
+        Number of clock edges from input deassertion to output deassertion
```
whitequark: This is really the number of flops, isn't it? Since the important part is MTBF.

nmigen/lib/cdc.py
```diff
@@ -1,7 +1,15 @@
...
 
-__all__ = ["MultiReg", "ResetSynchronizer"]
+__all__ = ["MultiReg", "ResetSynchronizer", "PulseSynchronizer", "Gearbox"]
+
+
+def _incr(signal, modulo):
+    if modulo == 2 ** len(signal):
+        return signal + 1
+    else:
+        return Mux(signal == modulo - 1, 0, signal + 1)
```
whitequark: There's already one in `lib.fifo` and this is getting ridiculous.

Wren6991: Agreed, it should be either inlined at my callsites, or preferably factored out. Wanted to keep the scope of this PR well-defined though.

nmigen/lib/cdc.py
Wren6991: You're right, MTBF is the important part here, delay is the job of CRG or some custom reset control logic. Will fix. Am away at the moment.

Wren6991:
Do you agree on my diagram of oMigen BusSynchronizer? I might have misunderstood e.g. the PulseSynchronizer logic. Haven't looked at it for a while now.

whitequark:
I'll need to think more about it as I wasn't the one who wrote it. @sbourdeauducq ?

sbourdeauducq:
> I've fixed the race by removing one stage from the datapath `MultiReg`.

But then, how do you preserve the MultiReg constraints between the anti-metastabilty registers?

Sounds easier to add one register of delay into the request path.

Wren6991:
@sbourdeauducq You're 100% right -- it's possible to just falsepath/MCP the connection between write data buffer and the MultiReg, but it's cleaner if it can just rely on constraints on paths internal to the MultiReg.

Added a patch which restores the data path to its original length, and lengthens the request path to compensate.

Wren6991:
Fixed merge conflict and made classes derive from Elaboratable (sorry for force-push)

nmigen/lib/cdc.py
Wren6991: Can you think of a nicer place to put this? It's a common pattern (as is saturating increment/decrement)

whitequark:
Thanks, I'll take another look soon.

Wren6991:
Updated naming conventions according to #97. I've also updated the only use of `lib.cdc` elsewhere in the standard library (in `lib.fifo`) to reflect this. Sorry for being unresponsive, day job has been busy

whitequark:
That's fine, I have a lot of other work here, so you're not really blocking anything.

Wren6991:
I saw this in #113

> CDC improvements like CDC safety (#4) and automatic false path constraints (#87; depends on #88). These are definitely needed, but there are still plenty of designs that will be just fine without them, and I think for most people waiting to pick up nMigen they wouldn't be an issue. Again, these are more of 0.2 issues.

And have also noted #87. Wanted to mention that falsepath is not necessarily safe for all of these primitives: in particular, BusSynchronizer is unsafe if there is too much skew between the request path and the datapath, and false paths can accumulate huge routing delays when FPGA utilisation is high. 

Since we want the widest possible toolchain support, one option is a maxdelay constraint of capture period minus setup. This isn't perfectly safe (clock skew), and does make closure harder, but might be worth considering.

whitequark:
> Wanted to mention that falsepath is not necessarily safe for all of these primitives

That seems very troublesome, and certainly something we want to loudly warn about. Would you be able to prepare a detailed evaluation of all of our CDC primitives with this in mind?

> BusSynchronizer is unsafe if there is too much skew between the request path and the datapath, and false paths can accumulate huge routing delays when FPGA utilisation is high.

Are there any ways to work around this?

> Since we want the widest possible toolchain support, one option is a maxdelay constraint of capture period minus setup. This isn't perfectly safe (clock skew), and does make closure harder, but might be worth considering.

This seems like a dramatic increase in complexity of platform code. It's an option, but maybe we can avoid it completely...

whitequark:
Relevant: https://github.com/m-labs/migen/commit/f4979a2403316942909bdfdd77e2a174a2b77857#commitcomment-34195474

Wren6991:
> Relevant: [m-labs/migen@f4979a2#commitcomment-34195474](https://github.com/m-labs/migen/commit/f4979a2403316942909bdfdd77e2a174a2b77857#commitcomment-34195474)

That is an interesting handshake, but seems to overlap quite a lot with BusSynchronizer. I have had the occasional shower thought that BusSynchronizer should have an optional valid/ready handshake in the launching domain (and perhaps a valid pulse in the capturing domain), which would give you something similar to this BlindSynchronizer.

I'm also happy to just port BlindSynchronizer wholesale!

sbourdeauducq:
I'm OK with merging them, but ``BlindTransfer`` is also commonly used without data a payload. (A typical use case for it is error reporting, and for many errors a simple pulse that tells you that the error occurred is enough)

whitequark:
@Wren6991 I've finally resolved #97; can you please update this PR?

whitequark:
Moving to 0.2, which has the theme of "standard library improvements" anyway.

nmigen/lib/cdc.py
```diff
@@ -42,17 +57,19 @@ class MultiReg(Elaboratable):
...
     - You are targeting an ASIC, or an FPGA that does not allow arbitrary initial flip-flop states;
-    - Your design features warm (non-power-on) resets of ``odomain``, so the one-time
+    - Your design features warm (non-power-on) resets of ``cd_o``, so the one-time
       initialization at power on is insufficient;
     - Your design features a sequenced reset, and the MultiReg must maintain its reset value until
-      ``odomain`` reset specifically is deasserted.
+      ``cd_o`` reset specifically is deasserted.
 
-    MultiReg is reset by the ``odomain`` reset only.
+    MultiReg is reset by the ``cd_o`` reset only.
```
madebr, line 65: `MultiReg` is renamed to `FFSynchronizer` in 8deb13cea3b0d51a974b1f5290a34736efe14558.

Issue #37, opened 2019-02-27
Please consider an optional reset on MultiReg


Wren6991:
I was browsing the code to learn the standard library, and bumped into this in cdc.py:

```
class MultiReg:
    def __init__(self, i, o, odomain="sync", n=2, reset=0):

     ...

        self._regs = [Signal(self.i.shape(), name="cdc{}".format(i),
                             reset=reset, reset_less=True, attrs={"no_retiming": True})
                      for i in range(n)]
```

I understand there are caveats with resets on CDC capture flops, but sometimes you need it. For example, a pulse-to-pulse synchroniser (or toggle synchroniser) contains a toggling NRZ signal crossing the domains, and the capture flops need a capture-domain reset to avoid spurious toggling observed by the logic when deasserting reset.

`platform.get_multi_reg(self)` may contain considerable magic for some synthesis flows, and should only need writing once; ideally, something like a pulse-to-pulse synchroniser should be implementable using the existing MultiReg.

A ligh-touch option could be to just add a `reset_less=True` to `MultiReg.__init__`, and pass this down.

Happy to be argued with, told to go implement it, or just told I'm wrong :)

Wren6991:
On a slightly-related note, are there any plans to flesh out the CDC library with e.g. pulse-to-pulse, Gray code, or ready-ack + multibit? Batteries included!

whitequark:
> I understand there are caveats with resets on CDC capture flops, but sometimes you need it. For example, a pulse-to-pulse synchroniser (or toggle synchroniser) contains a toggling NRZ signal crossing the domains, and the capture flops need a capture-domain reset to avoid spurious toggling observed by the logic when deasserting reset.

Weak yes. I will need to dig in and understand the context here much better. You can help this by providnig compelling real-world examples and extensive documentation. I.e. teach me the use cases.

> On a slightly-related note, are there any plans to flesh out the CDC library with e.g. pulse-to-pulse, Gray code, or ready-ack + multibit? Batteries included!

Yes, all this was in Migen and it will get ported. Gray counters are already in.

Wren6991:
> Weak yes. I will need to dig in and understand the context here much better. You can help this by providnig compelling real-world examples and extensive documentation. I.e. teach me the use cases.

Okay I will try! I don't guarantee any of the below code works, but it's conceptually sound

[...]

PR #44, opened 2019-03-19
lib.cdc: add optional reset to MultiReg, and document its use cases

Wren6991:
Following on from #37.

This patch adds a `reset_less` parameter to MultiReg. The default behaviour is the same as the original behaviour. It also adds a docstring to MultiReg which, besides usual docstring stuff, adds the following footnote:

```
    Note on Reset
    -------------
    MultiReg is non-resettable by default. Usually this is the safest option;
    on FPGA the MultiReg will still be initialised to its `reset` value,
    when the FPGA loads its configuration bitstream.

    However, in designs where the value of the MultiReg must be valid
    immediately after reset, consider `reset_less`=False if:
     - You are targeting an ASIC, or an FPGA that does not initialise
       flipflop values
     - Your design features warm (non-power-on) resets of `odomain`, so
       the one-time initialisation at power on is insufficient
     - Your design features a sequenced reset, and the MultiReg must maintain
       its reset value until `odomain` reset specifically is deasserted.

    MultiRegs are reset by the `odomain` reset only.
```

I made sure to make it explicit that the default behaviour relies on FPGA bitstream config. Not sure if this is exactly the kind of documentation you intended -- let me know and I will fix it up :)

Thanks

### DIFF
--- a/nmigen/lib/cdc.py
+++ b/nmigen/lib/cdc.py
@@ -5,13 +5,60 @@ __all__ = ["MultiReg", "ResetSynchronizer"]
 
 
 class MultiReg:
-    def __init__(self, i, o, odomain="sync", n=2, reset=0):
+    """Resynchronise a signal to a different clock domain.
+
+    Consists of a chain of flipflops. Eliminates metastabilities
+    at the output, but provides no other guarantee as to the safe
+    domain-crossing of a signal.
+
+    Parameters
+    ----------
+    i : Signal(), in
+        Signal to be resynchronised
+    o : Signal(), out
+        Signal connected to synchroniser output
+    odomain : str
+        Name of output clock domain
+    n : int
+        Number of flops between input and output.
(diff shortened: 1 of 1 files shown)


whitequark:
Thank you! I've manually merged this PR, with the following changes:
 
  * Some typos (`int` instead of `bool`, etc).
  * In ReST, single backticks are used for emphasis, double backticks are used for code/monospace.
  * Clarified some details in the note on reset (all FPGAs allow initializing flip-flops to at least zero; any of the listed conditions can be true for the reset to become necessary).
  * Reflowed the text for 100 columns, which is what's used in this codebase.

Wren6991:
> int instead of bool, etc

Oops, potentially bad C codebase at work leaking through.

Thanks for merging, and sorry for the mess; that's great feedback. I'll go and rebase + fix up my other PR based on this.

PR #45, opened 2019-03-19
lib.io: import CRG from Migen.

jfng:
Tests are missing.

### DIFF
--- a/nmigen/lib/io.py
+++ b/nmigen/lib/io.py
@@ -2,7 +2,7 @@ from .. import *
 from ..hdl.rec import *
 
 
-__all__ = ["pin_layout", "Pin"]
+__all__ = ["pin_layout", "Pin", "CRG"]
 
 
 def pin_layout(width, dir, xdr=1):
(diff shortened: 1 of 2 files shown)


whitequark:
I am wondering if the nMigen CRG implementation should take into account FPGAs like iCE40, where the BRAMs cannot be accessed for some time after startup. (This is very poorly documented but it's a hard requirement...)

So, my CRG for iCE40 uses a timer.

jfng:
> I am wondering if the nMigen CRG implementation should take into account FPGAs like iCE40, where the BRAMs cannot be accessed for some time after startup. (This is very poorly documented but it's a hard requirement...)

Is this what you meant ?

whitequark:
Yes. Isn't it sufficient to use ResetInserter here?

whitequark:
I think there is no particular need for a generic CRG in nMigen because various `vendor.*` packages use highly platform-specific reset generation and clock gating gateware which is much better at this job.

PR #46, opened 2019-03-19
[WIP] Add nmigen.build

jfng:
* Tests are missing.
* This PR depends on https://github.com/m-labs/nmigen/pull/45

This is a proof-of-concept for a nMigen equivalent of Migen's build system.

The main difference lies in the use of [Edalize](https://github.com/olofk/edalize) to invoke vendor tools instead of doing so ourselves.

However, as far as I understand, constraint file (XDC, PCF, etc.) formatting must still be done on our side, which prevents nMigen from being completely vendor agnostic.

### DIFF
--- a/nmigen/back/rtlil.py
+++ b/nmigen/back/rtlil.py
@@ -848,8 +848,11 @@ def convert_fragment(builder, fragment, name, top):
     return module.name, port_map
 
 
-def convert(fragment, name="top", **kwargs):
-    fragment = ir.Fragment.get(fragment, platform=None).prepare(**kwargs)
+def convert(fragment, name="top", platform=None, **kwargs):
+    fragment = ir.Fragment.get(fragment, platform)
+    if platform is not None:
+        kwargs['ports'] = kwargs.get('ports', []) + platform.ports
+    fragment = fragment.prepare(**kwargs)
     builder = _Builder()
     convert_fragment(builder, fragment, name=name, top=True)
     return str(builder)
--- a/nmigen/build/generic_platform.py
+++ b/nmigen/build/generic_platform.py
@@ -1,0 +1,383 @@
+import edalize
+import os
+from collections import defaultdict, OrderedDict
+
+from .. import *
+from ..back import verilog
+from ..hdl.rec import *
+from ..lib.io import *
+
+
+__all__ = [
+    "Pins", "Subsignal", "DiffPairs", "Resource", "Connector",
+    "ConstraintError", "GenericPlatform"
+]
+
+
+class Pins:
+    def __init__(self, *identifiers, dir="io"):
+        for i in identifiers:
+            if not isinstance(i, str):
+                raise TypeError("Identifier should be a string, not {!r}".format(i))
+        self.identifiers = " ".join(identifiers).split(" ")
+
+        if dir not in ("i", "o", "io"):
+            raise TypeError("Direction should be either 'i', 'o' or 'io', not {!r}"
+                            .format(dir))
+        self.dir = dir
+
+    def __len__(self):
+        return len(self.identifiers)
+
+    def __repr__(self):
+        return "(pins {} '{}')".format(" ".join(self.identifiers), self.dir)
+
+
+class DiffPairs:
+    def __init__(self, *pairs, dir="io"):
+        p, n = zip(*(pair.split(" ") for pair in pairs))
+        self.p = Pins(*p, dir=dir)
+        self.n = Pins(*n, dir=dir)
+
+        self.dir = self.p.dir
+
+    def __len__(self):
+        return len(self.p)
+
+    def __repr__(self):
+        return "(diffpairs {} {})".format(self.p, self.n)
+
+
+class Subsignal:
+    def __init__(self, name, *io, extras=None):
+        self.name = name
+
+        self.has_pins       = len(io) == 1 and isinstance(io[0], Pins)
+        self.has_diffpairs  = len(io) == 1 and isinstance(io[0], DiffPairs)
+        self.has_subsignals = io and all(isinstance(c, Subsignal) for c in io)
+
+        if not (self.has_pins or self.has_diffpairs or self.has_subsignals):
+            raise TypeError("I/O constraints should be either Pins or DiffPairs or "
+                            "multiple Subsignal, not {!r}"
+                            .format(io))
+        self.io = io
+
+        extras = extras or []
+        for c in extras:
+            if not isinstance(c, str):
+                raise TypeError("Extra contraint should be a string, not {!r}".format(c))
+        self.extras = list(extras)
+
+        if self.has_subsignals:
+            for sub in self.io:
+                sub.extras += self.extras
+
+    def __repr__(self):
+        return "(subsignal {} {} {})".format(self.name, " ".join(map(repr, self.io)), self.extras)
+
+
+class Resource(Subsignal):
+    def __init__(self, name, number, *io, extras=None):
+        self.number = number
+        super().__init__(name, *io, extras=extras)
+
+    def __repr__(self):
+        return "(resource {} {} {} {})".format(self.name, self.number, " ".join(map(repr, self.io)), self.extras)
+
+
+class Connector:
+    def __init__(self, name, description):
+        self.name = name
+
+        if isinstance(description, str):
+            self.description = dict(enumerate(description.split(" ")))
+        elif isinstance(description, list):
+            self.description = dict(enumerate(description))
+        elif isinstance(description, dict):
+            self.description = description
+        else:
+            raise TypeError("Description should be either a whitespace-separated string, "
+                            "a list or a dict, not {!r}"
+                            .format(description))
+
+        for alias, identifier in self.description.items():
+            if not isinstance(identifier, str):
+                raise TypeError("Identifier should be a string, not {!r}".format(identifier))
+            elif identifier == "None":
+                self.description[alias] = None
+
+    def __getitem__(self, name):
+        return self.description[name]
+
+    def __repr__(self):
+        return "(connector {} {})".format(self.name, self.description)
+
+
+def _yosys_id_escape(s):
+    if s[0] == "\\":
+        s = s[1:]
+    # TODO escape verilog keywords
+    if s[0].isdigit() or not s.replace("_", "").isalnum():
+        s = "\\{} ".format(s)
+    return s
+
+
+class ConstraintError(Exception):
+    pass
+
+
+class _ConstraintManager:
+    def __init__(self, resources, connectors):
+        self.available    = defaultdict(dict)
+        self.matched      = OrderedDict()
+        self.connectors   = dict()
+        self.commands     = []
+        self.clocks       = OrderedDict()
+        self.constraints  = OrderedDict()
+        self.tristates    = []
+        self.diffpairs    = []
+        self.ports        = []
+
+        self.add_resources(resources)
+        self.add_connectors(connectors)
+
+    def add_resources(self, resources):
+        for r in resources:
+            if not isinstance(r, Resource):
+                raise TypeError("Object {!r} is not a Resource".format(r))
+            if r.number in self.available[r.name]:
+                raise NameError("Resource {!r} has a (name, number) pair that is already taken"
+                                .format(r))
+            self.available[r.name][r.number] = r
+
+    def add_connectors(self, connectors):
+        for c in connectors:
+            if not isinstance(c, Connector):
+                raise TypeError("Object {!r} is not a Connector".format(c))
+            if c.name in self.connectors:
+                raise NameError("Connector {!r} has a name that is already taken".format(c))
+            self.connectors[c.name] = c
+
+    def add_command(self, command):
+        if not isinstance(command, str):
+            raise TypeError("Command must be a string, not {!r}".format(command))
+        self.commands.append(command)
+
+    def add_period_constraint(self, resource, period):
+        if not isinstance(resource, Resource):
+            raise TypeError("Object {!r} is not a Resource".format(resource))
+        if (resource.name, resource.number) in self.clocks:
+            raise ConstraintError("Resource {!r} is already constrained to a period"
+                                  .format(resource))
+        self.clocks[resource.name, resource.number] = period
+
+    def lookup(self, name, number=None):
+        if number is None and self.available[name]:
+            number = sorted(self.available[name])[0]
+        elif number not in self.available[name]:
+            raise NameError("No available Resource with name '{}' and number '{}'"
+                            .format(name, number))
+        return self.available[name][number]
+
+    def request_raw(self, name, number=None):
+        resource = self.lookup(name, number=number)
+        assert (resource.name, resource.number) not in self.matched
+
+        def get_shape(subsignal):
+            if subsignal.has_subsignals:
+                return Layout([(s.name, get_shape(s)) for s in subsignal.io])
+            elif subsignal.has_diffpairs:
+                pair = subsignal.io[0]
+                return Layout([("p", len(pair.p)), ("n", len(pair.n))])
+            else:
+                return len(subsignal.io[0])
+
+        port_shape = get_shape(resource)
+        port_name = "{}_{}".format(resource.name, resource.number)
+
+        if isinstance(port_shape, Layout):
+            port = Record(port_shape, name=port_name)
+        else:
+            port = Signal(port_shape, name=port_name)
+
+        def add_constraints(port, subsignal):
+            if subsignal.has_subsignals:
+                for s in subsignal.io:
+                    add_constraints(port[s.name], s)
+            elif subsignal.has_diffpairs:
+                self.constraints[_yosys_id_escape(port.p.name)] = subsignal.io[0].p, subsignal.extras
+                self.constraints[_yosys_id_escape(port.n.name)] = subsignal.io[0].n, subsignal.extras
+                self.ports += port.p, port.n
+            else:
+                self.constraints[_yosys_id_escape(port.name)] = subsignal.io[0], subsignal.extras
+                self.ports.append(port)
+
+        add_constraints(port, resource)
+
+        del self.available[resource.name][resource.number]
+        self.matched[resource.name, resource.number] = port
+        return port
+
+    def request(self, name, number=None, dir=None):
+        resource = self.lookup(name, number=number)
+        assert (resource.name, resource.number) not in self.matched
+
+        def resolve_dir(subsignal, dir):
+            if subsignal.has_subsignals:
+                dir = dir or dict()
+                if not isinstance(dir, dict):
+                    raise TypeError("{} has subsignals. Directions should be a dict, not {!r}"
+                                    .format(subsignal.name, dir))
+                for s in subsignal.io:
+                    dir[s.name] = resolve_dir(s, dir.get(s.name, None))
+            else:
+                dir = dir or subsignal.io[0].dir
+                if dir not in ("i", "o", "io"):
+                    raise TypeError("Direction should be either 'i', 'o' or 'io', not {!r}"
+                                    .format(dir))
+                if subsignal.io[0].dir != "io" and dir != subsignal.io[0].dir:
+                    raise ValueError("Direction {} cannot be changed to {}. Valid changes are "
+                                     "'io'->'i', 'io'->'o'."
+                                     .format(subsignal.io[0].dir, dir))
+            return dir
+
+        dir = resolve_dir(resource, dir)
+
+        def get_shape(subsignal, dir):
+            if subsignal.has_subsignals:
+                return Layout((s.name, get_shape(s, dir[s.name])) for s in subsignal.io)
+            else:
+                shape = len(subsignal.io[0])
+                if dir == "io":
+                    shape = Layout([("o", shape), ("oe", 1), ("i", shape)])
+                return shape
+
+        port_shape = get_shape(resource, dir)
+        port_name = "{}_{}".format(resource.name, resource.number)
+
+        if isinstance(port_shape, Layout):
+            port = Record(port_shape, name=port_name)
+        else:
+            port = Signal(port_shape, name=port_name)
+
+        def add_constraints(port, subsignal, dir):
+            if subsignal.has_subsignals:
+                for s in subsignal.io:
+                    add_constraints(port[s.name], s, dir[s.name])
+            elif subsignal.has_diffpairs:
+                p = Signal.like(port, name=port.name + "_p")
+                n = Signal.like(port, name=port.name + "_n")
+                self.diffpairs.append((port, p, n, dir))
+                self.constraints[_yosys_id_escape(p.name)] = subsignal.io[0].p, subsignal.extras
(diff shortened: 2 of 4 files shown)


nmigen/back/rtlil.py
```diff
@@ -843,13 +843,17 @@ def convert_fragment(builder, fragment, name, top):
     # to create a cell for us in the parent module.
     port_map = OrderedDict()
     for signal in fragment.ports:
-        port_map[compiler_state.resolve_curr(signal)] = signal
+        signal.name = compiler_state.resolve_curr(signal)
+        port_map[signal.name] = signal
```
whitequark: That's backwards. The constraint file should be adapted to the signal names, not the other way around.

whitequark:
What thought has been put into the design of `nmigen.build`? Where are the design notes? What problems with `migen.build` does it fix? I don't see any of this...

mithro:
FYI - @olofk

sbourdeauducq:
> The main difference lies in the use of [Edalize](https://github.com/olofk/edalize) to invoke vendor tools instead of doing so ourselves.

I want to invoke Vivado myself from Nix to facilitate things like https://github.com/m-labs/nix-scripts/issues/11.
What problem does edalize solve?

olofk:
@sbourdeauducq Edalize is an abstraction library for EDA tools to make it a bit easier to target different tools. It does not attempt to generate any RTL or constraint files as the purpose it to be a thin layer which allows python libraries (e.g. migen, litex, myhdl, apio) or stand-alone applications to use it while leaving the details up to each of these projects.

For the purpose of nmigen we have identified a couple of things we would like to change to make edalize more in lin with nmigens goals. IIRC the most important of those was to generate .sh/.bat files instead of Makefiles, to avoid the need to have `make` on the target system. I did a quick PoC for the icestorm backend some time ago (https://github.com/olofk/edalize/commits/sh) but haven't found time to finish this work

sbourdeauducq:
Why is it better than ``subprocess.Popen(["vivado", "-mode", ...], ...)``?

olofk:
It allows rebuilding the project without having python installed. It's of course a trade-off, but I have found it very handy to export a system that I can send to clients, or to a build server, without any external dependencies.

With that said, I'm not opposed to adding a possibility to launch tools from within the python code. E.g. by adding a a constructor flag to choose between generating sh/bat files, makefiles or launch from python. But I also got the feeling from your previous message that you wanted to launch vivado yourself anyway, so in that case, it would probably be best to just generate the sh file, which you can ignore and launch vivado yourself

sbourdeauducq:
> It allows rebuilding the project without having python installed. It's of course a trade-off, but I have found it very handy to export a system that I can send to clients, or to a build server, without any external dependencies.

The existing solution in Migen also allows this. And Nix can *automatically* send stuff to one or many build servers...

olofk:
> > It allows rebuilding the project without having python installed. It's of course a trade-off, but I have found it very handy to export a system that I can send to clients, or to a build server, without any external dependencies.
> 
> The existing solution in Migen also allows this. And Nix can _automatically_ send stuff to one or many build servers...

Not sure what _this_ refers to, but if that was a reply to what I wrote about sending to build servers, then good you have a solution for that, because that is outside of edalize's scope. If _this_ refers to rebuilding an EDA project without requiring python, then it seems you had the answer to your question already.

All in all, I don't have any grand plans for edalize other than being a (mostly) tool-agnostic interface to EDA tools that can be reused by other projects to quickly gain support for many EDA tools. If it fits nmigen, great. More users is generally a good thing. If it needs some changes, let's look at those changes. If it's a bad fit for nmigen, then there's no use in shoehorning it in.

sbourdeauducq:
> If _this_ refers to rebuilding an EDA project without requiring python, then it seems you had the answer to your question already.

It refers to building without python. That has been used several times to report problems to Xilinx.

>  If it fits nmigen, great. More users is generally a good thing.

Well it fits in the sense that it can do the equivalent of a few ``subprocess.Popen`` and other calls, but I don't see it solving a difficult problem. And I see the future synthesis flows mostly using Nix to call the EDA programs and not being based on all-in-one Python scripts. The latter are fine for simple projects though.

sbourdeauducq:
> to quickly gain support for many EDA tools

The difficult part here is writing the tcl and xdc files (or their equivalents for other tools)...

jfng:
### Design goals

1. Equivalent functionality to migen.build
2. Easy migration of Migen platforms to nMigen
3. Maintain portability between Linux/macOS/Windows

### Issues with migen.build

* Interfacing with vendor toolchains requires a lot of OS-specific code to format paths, file extensions, build scripts, etc. See [lattice/diamond.py](https://github.com/m-labs/migen/blob/master/migen/build/lattice/diamond.py#L90-L124), [lattice/icestorm.py](https://github.com/m-labs/migen/blob/master/migen/build/lattice/icestorm.py#L36-L49).

Externalizing toolchain invocation to another tool/library could alleviate the need for this code.

### Issues with Edalize

* Requiring Make to build designs is a concern for portability.
* Constraints file formatting must be done by nMigen anyway.

### Current approach and limitations

So far, the approach for this PoC has been to reuse as much code from Migen as possible, except for toolchain-specific code, which is now limited to constraints file formatting.

As the io/connector handling code remains unchanged, most platforms should hopefully be able to be migrated with little to no modifications.

In the case of Edalize+Vivado, I do not know how to implement hooks such as `toolchain.additional_commands`, which can be found in [platforms/arty_a7.py](https://github.com/m-labs/migen/blob/master/migen/build/platforms/arty_a7.py#L241-L243).

This is a case were we need to add commands to be executed after `write_bitstream`, and the Edalize [template](https://github.com/olofk/edalize/blob/master/edalize/templates/vivado/vivado-run.tcl.j2) is currently not extensible.
This is also an example of a platform which requires toolchain configuration *outside* the constraints file.

I am not sure if my current approach is the right one and I am ready to turn backward and add toolchain invocation inside nMigen if need be.

sbourdeauducq:
>     3\. Maintain portability between Linux/macOS/Windows

Except for the open source flow, I don't think anything runs on Mac.

sbourdeauducq:
> * Equivalent functionality to migen.build
> 
> * Easy migration of Migen platforms to nMigen

If the goal is to do exactly the same as Migen, there is no need for nMigen :)

The platform migration can probably be done with some script that reads the Migen platform file and generates a rough nMigen platform file, that can be polished by hand afterwards.

olofk:
> In the case of Edalize+Vivado, I do not know how to implement hooks such as `toolchain.additional_commands`, which can be found in [platforms/arty_a7.py](https://github.com/m-labs/migen/blob/master/migen/build/platforms/arty_a7.py#L241-L243).
> 
> This is a case were we need to add commands to be executed after `write_bitstream`, and the Edalize [template](https://github.com/olofk/edalize/blob/master/edalize/templates/vivado/vivado-run.tcl.j2) is currently not extensible.

All Edalize backends that support tcl (currently ise, modelsim, quartus, rivierapro, spyglass and vivado) can be extended by passing tcl files and specifying them as tclSource. The backend will then source these files as part of the project. This can be used to set extra options or install hooks. The other option is to implement this as a [post_build hook](https://edalize.readthedocs.io/en/latest/edam/api.html#hook)

> This is also an example of a platform which requires toolchain configuration _outside_ the constraints file.
> 

Can you explain what kind of configuration you're referring too?

jfng:
> Can you explain what kind of configuration you're referring too?

If I'm not mistaken, in the previous example:
`write_cfgmem -force -format bin -interface spix4 -size 16 -loadbit \"up 0x0 {build_name}.bit\" -file {build_name}.bin"`
should end up in a tcl file separate from the xdc ?

whitequark:
I don't think that *requiring* Nix for nMigen is the way to go. For example, Glasgow supports Windows as a 1st class platform, which means that it could not use nMigen; I see no reason nMigen should specifically eschew Windows given that it does not really gain much from this.

I think Edalize generating shell or batch scripts with an appropriate abstraction is a good solution to the problem with `migen.build` where it would poorly construct ad-hoc build scripts with a lot of code duplication.

I also think it's more important to focus on problems with `migen.build` such as the pin management system as opposed to bikeshedding the best way to run a few subprocesses.

whitequark:
@jfng I apologize for my tone earlier in this thread (in https://github.com/m-labs/nmigen/pull/46#issuecomment-474535969). I see you understood me correctly, but I should not have implied that you haven't done design work.

sbourdeauducq:
It's not really bikeshedding since it leads to the idea that nmigen.build (for lack of a better name) should focus on managing FPGA I/O resources and generating EDA tool inputs, and not try too hard to be a build system. The user can supply their own external build system, e.g. based on Nix, that runs the EDA tools and performs other steps.

Also, the generated EDA tool inputs should not be monolithic as they are now. For example:
* the user may want to generate a Vivado DCP and not a final bitstream (which the external build system will put through additional steps before generating the bitstream itself)
* the external build system may supply itself the Verilog/VHDL files for additional cores (LM32, mor1kx, VexRiscv, Xilinx IP)
* having an external build system provides additional flexibility and features to mix in SpinalHDL or Xilinx IP generator flows.

Having a "batteries included" script that runs all common steps and invokes the EDA tools is fine for simple projects though.

whitequark:
> and not try too hard to be a build system

I agree, which is why I think Edalize is the way to go.

> The user can supply their own external build system, e.g. based on Nix, that runs the EDA tools and performs other steps.

> Also, the generated EDA tool inputs should not be monolithic as they are now.

Also reasonable.

cr1901:
@jfng Re: edalize being tied to make, the plan was to move away from that. @olofk even has a branch for it.

Re: "external build system", is that not what edalize is already doing? It will also generate the EDA inputs for you, so nMigen doesn't have to do it. There should be a "fire and forget" way to generate all the EDA inputs into a self-contained package when using nMigen, but I'm not even convinced nMigen itself needs to generate those files. NMigen can "just" focus on pin and resource management.

sbourdeauducq:
> Re: "external build system", is that not what edalize is already doing? It will also generate the EDA inputs for you, so nMigen doesn't have to do it.

No to both. Edalize basically just runs the tool; and for the purposes of things like https://github.com/m-labs/nix-scripts/issues/11 I want to be able to disable the Edalize steps completely.

cr1901:
@sbourdeauducq You can use edalize to just generate a self contained project. See [here](https://github.com/cr1901/crispy-succotash/blob/nmigen/baseboard.py#L396) for an example.

Anyway is below an accurate summary of what you're discussing?

* Your nix scripts in effect take care of the boilerplate such as generating e.g. the EDA .tcl input files that customize the build steps. In your scenario, Migen only generates the UCF?
* You are open to having nMigen generate EDA input files in addition to a UCF, which you then run directly from your shell rather than from Python subprocesses.
* In all cases, the added functionality should not interfere with your Nix flow.

The advantage of migen.build for me was mostly reducing boilerplate (TCL, CDC, etc) for every project I wrote. Yes there was a lot of duplication internally in migen.build, but over the years migen.build saved me a lot of time.

sbourdeauducq:
> You can use edalize to just generate a self contained project. See [here](https://github.com/cr1901/crispy-succotash/blob/nmigen/baseboard.py#L396) for an example

It still generates a monolithic output, doesn't allow an external build system to supply additional VHDL/Verilog files, doesn't allow generating a DCP instead of a bitfile, etc.

Again, having the option to do that, as long as it can be disabled, is fine and good for simple projects.

jfng:
> some constraint names are Xilinx-centric (e.g. IOStandard). Do we need any constraint translation layer anyway or just put everything into something like the current Misc?

I agree to put vendor specific constraints in `Misc`.
What is the purpose of the `PlatformInfo` constraint ? I do not see it used in Migen platforms.

If `IOStandard`, `Drive` and `PlatformInfo` are not required, we could end up with only three types of constraint: `Pins`, `Subsignal` and `Misc`.

> * error reporting
> * code cleanliness in the pin management system

I think these two issues are related.
In migen.build, an I/O resource is defined by a name, a number and a list of constraints.

As far as I understand, the latter are subject to the following rules:
1. A resource must either have one `Pins` constraint or at least one `Subsignal` constraint.
2. A `Subsignal` constraint must contain a `Pins` constraint.

These rules are implicitly validated in the [_resource_type()](https://github.com/m-labs/migen/blob/master/migen/build/generic_platform.py#L76-L95) function. However, violations are reported by an abrupt `AssertionError` with no further indication.

But despite being treated differently from other constraints, `Pins` and `Subsignal`s can be positioned anywhere in the constraint list. This lack of distinction leads to intricate code where functions have to do the separation themselves. See [get_sig_constraints()](https://github.com/m-labs/migen/blob/master/migen/build/generic_platform.py#L201-L228).

In order to explicit this distinction, I suggest the following model:

An I/O resource is defined by:
* a name,
* a number,
* either one `Pins` constraint or a list of `Subsignal` constraints,
* an optional list of `Misc` constraints.

A `Subsignal` is defined by:
* a name,
* a `Pins` constraint,
* an optional list of `Misc` constraints.

By separating `Pins` and `Subsignal`s from `Misc` constraints, resource validation and retrieval should become straighforward.

> > Re: "external build system", is that not what edalize is already doing? It will also generate the EDA inputs for you, so nMigen doesn't have to do it.
> 
> No to both. Edalize basically just runs the tool; and for the purposes of things like [m-labs/nix-scripts#11](https://github.com/m-labs/nix-scripts/issues/11) I want to be able to disable the Edalize steps completely.

We can split the build in two steps:
1. **Toolchain input generation**: the Verilog output, constraints file, additional TCL files, etc. are generated by nMigen.
2. **Build**: files generated from step 1. are fed to Edalize to produce a bitstream.

Step 2. could then be made optional for use cases where the user wants more control over their toolchain.

jfng:
> In order to make this distinction explicit, I suggest the following model:
> 
> An I/O resource is defined by:
>     * a name,
>     * a number,
>     * either one `Pins` constraint or a list of `Subsignal` constraints,
>     * an optional list of `Misc` constraints.
> 
> A `Subsignal` is defined by:
>     * a name,
>     * a `Pins` constraint,
>     * an optional list of `Misc` constraints.
> 
> By separating `Pins` and `Subsignal`s from `Misc` constraints, resource validation and retrieval should become straightforward.

For example:

Before:
```python
_io = [
    ("clk100", 0, Pins("E3"), IOStandard("LVCMOS33")),
    ("serial", 0,
        Subsignal("tx", Pins("D10")),
        Subsignal("rx", Pins("A10")),
        IOStandard("LVCMOS33")
    ),
]
```

After:
```python
_io = [
    Resource("clk100", 0, Pins("E3"), misc=[Misc("IOSTANDARD=LVCMOS33")]),
    Resource("serial", 0,
        Subsignal("tx", Pins("D10")),
        Subsignal("rx", Pins("A10"))
        misc=[Misc("IOSTANDARD=LVCMOS33")]
    )
]

class Resource:
    def __init__(self, name, number, *io, misc=[]):
        self.name = name
        self.number = number

        if isinstance(io[0], Pins) and len(io) == 1:
            self.io = io[0]
        elif all(isinstance(c, Subsignal) for c in io):
            self.io = io
        else:
            raise TypeError(f"io has invalid type: should be either Pins or a list of Subsignals")

        if not all(isinstance(c, Misc) for c in misc):
            raise TypeError(f"misc has invalid type: should be a list of Misc")
        self.misc = misc
```

sbourdeauducq:
> ```python
>         if misc and not all(isinstance(c, Misc) for c in misc):
>             raise TypeError("misc has invalid type: should be a list of Misc")
> ```

What's the point of ``Misc`` then? Just make a list of strings. Also, the usual Python coding style is not to check for types unnecessarily.

whitequark:
> Also, the usual Python coding style is not to check for types unnecessarily.

That style is garbage because all it results in is crashes some time later where you no longer have any context. The hypothetical advantages of being able to pass an object that implements the `str` interface but isn't a string are never realized here as well.

sbourdeauducq:
I don't think those errors are very hard to debug, and you can still have a similar issue by putting something other than a string inside the ``Misc`` object anyway. Python is simply not meant to be used without duck typing.

whitequark:
> I don't think those errors are very hard to debug

I do, sufficiently to add type and range checks everywhere else in nMigen. Incidentally, almost everyone who gave me feedback on nMigen appreciates these checks.

> and you can still have a similar issue by putting something other than a string inside the `Misc` object anyway

Which is why the `Misc` object also needs one, yes.

> Python is simply not meant to be used without duck typing.

I completely disagree. Python doesn't even support duck typing, in the sense that it gives you essentially no tools to actually make use of that feature. For example, `ResetInserter` and co should be mostly transparent proxies, ideally, which is a single method in Ruby (`def method_missing`) but in Python I still don't see an especially good way to do that.

Also, Linux is simply not meant to be used on desktops, and yet here we are. Of course, I would rather not use Python here at all, or ever again. But so long as this is a requirement, I'm going to make it work reliably whether it's meant for that or not.

jfng:
> All Edalize backends that support tcl (currently ise, modelsim, quartus, rivierapro, spyglass and vivado) can be extended by passing tcl files and specifying them as tclSource. The backend will then source these files as part of the project. This can be used to set extra options or install hooks. The other option is to implement this as a [post_build hook](https://edalize.readthedocs.io/en/latest/edam/api.html#hook)

@olofk Still with the Vivado backend, if I wanted to do the following:
```tcl
synth_design -top {top} -part {device}
# ask for reports about utilization, timing, etc here
opt_design
place_design
# more reports
route_design
# more reports, checkpoint
write_bitstream {top}.bit
```
Can I get away with sourcing these commands in an additional file (or a "pre_build" hook), except `write_bitstream` which would  be executed by rendering the [vivado-run.tcl.j2](https://github.com/olofk/edalize/blob/fa590aba08449a3d6b01930f5aa56d833ba22c23/edalize/templates/vivado/vivado-run.tcl.j2#L4) template ?

jfng:
Progress update:
Changes:
* improve constraint error checking
* `Subsignal`s can now be nested

Todo:
* automatic insertion of differential buffers
* split build in two steps, as discussed previously

jfng:
Progress update:

> Todo:
> * automatic insertion of differential buffers

Automatic insertion of tristate buffers needs to be sorted out first.

Changes:
* `Pins` take an optional `dir` argument:
  * valid values are "i", "o" or "io" (default)
  * `dir="io"` tristates the pin

* `<platform>.request()` takes an optional `dir` argument which overrides the direction:
  * the direction can only be changed from "io" to "i" or "o"
  * in case of subsignals, `dir` must be a dict.
e.g. `platform.request("serial", dir={"tx": "o"})`

jfng:
Progress update:

* Support for differential pairs:
    * They are defined with `DiffPairs` instead of `Pins`
    * Identifiers are provided as "P N" pairs
    * Platforms must implement `.get_diff_input()`, `.get_diff_output()`, `.get_diff_tristate()` to return the buffer primitives.
    * Example usage:
```python3
Resource("foo", 0,
    DiffPairs(
        "N2 N1",
        "U2 V2",
        dir="io"
    ),
    misc=[Misc("IOSTANDARD=DIFF_SSTL135")]
)
```

* An ECP5 platform using Trellis+nextpnr has been added.

sbourdeauducq:
> ```python
>  misc=[Misc("IOSTANDARD=DIFF_SSTL135")]
> ```

i still don't see the point of encapsulating every string in a ``Misc`` instance, instead of just writing ``misc=["IOSTANDARD=DIFF_SSTL135"]``.

jfng:
> i still don't see the point of encapsulating every string in a `Misc` instance, instead of just writing `misc=["IOSTANDARD=DIFF_SSTL135"]`.

`misc=[Misc("..."), ...]` is now `extras=["...", "..."]`

nmigen/lib/io.py
```diff
@@ -45,3 +45,48 @@ def elaborate(self, platform):
...
+
+
+class DiffInput:
+    def __init__(self, p, n, i):
+        self.p = p
+        self.n = n
+        self.i = i
+
+    def elaborate(self, platform):
+        m = Module()
```
whitequark: The entire body of this function can be replaced with `platform.get_diff_input(self.p, self.n, self.i)`, it is a no-op as written.

But also I think we don't actually need any of these classes. You either:
  * request a prepared pad (SDR, DDR, whatever) from the platform, in which case it gives you a Record already connected to an Instance, or
  * request a raw pad from the platform, in which case it gives you a bare inout.

In both cases we don't need `DiffInput`, etc, only `pad_layout`.

nmigen/build/lattice.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def _format_lpf(platform, name):
+    text = ""
+    text += "BLOCK RESETPATHS;\n"
+    text += "BLOCK ASYNCPATHS;\n"
+
+    for port_name, identifiers, extras in platform.get_port_constraints():
+        if len(identifiers) == 1:
+            text += "LOCATE COMP \"{}\" SITE \"{}\";\n".format(port_name, identifiers[0])
```
whitequark: Please use raw strings (`r"{}"

nmigen/build/generic_platform.py
```diff
@@ -0,0 +1,374 @@
...
+            raise NameError("No available Resource with name '{}' and number '{}'"
+                            .format(name, number))
+        return self.available[name][number]
+
+    def request(self, name, number=None, dir=None, xdr=1):
+        resource = self.lookup(name, number=number)
+        assert (resource.name, resource.number) not in self.matched
+
+        if dir is not None:
+            resource.update_dir(dir)
```
whitequark: How are resources supposed to be added? If this is done the same way as in oMigen, i.e. you have a global array of `_io` somewhere, then this is completely wrong, because trying to build two different designs in one Python invocation will fail.

In general, nMigen avoids mutating anything, which is slightly slower but eliminates hard-to-find bugs like this.

jfng:
Changes:

* Overriding `dir` in `<platform>.request()` does not mutate global state anymore,
* `<platform>.request_raw()` can be used to request a bare resource,
* `DiffInput` and friends have been removed.

nmigen/build/generic_platform.py
```diff
@@ -0,0 +1,383 @@
+import edalize
```
sbourdeauducq: Can we do a lazy import so it's not a dependency for flows not using edalize?

whitequark: We've agreed (privately on IRC) to postpone edalize integration as edalize does not yet do what we want. I'm not sure why the PR isn't updated.

whitequark:
Merged.

whitequark:
@olofk By the way, you might want to investigate the solution I made for nMigen instead of Edalize:

  * https://github.com/m-labs/nmigen/blob/3194b5c90ba8bc22b600e25ab619dcac2582d476/nmigen/build/run.py#L1-L62
  * https://github.com/m-labs/nmigen/blob/3194b5c90ba8bc22b600e25ab619dcac2582d476/nmigen/build/plat.py#L179-L268
  * https://github.com/m-labs/nmigen/blob/master/nmigen/vendor/fpga/lattice_ice40.py#L16-L109

The way it works is it uses Jinja2 templates to generate a *completely standalone* archive of build precursors, and scripts for all supported OSes, which you can then freely move around (as an archive or in any other way) and build by running one command. As far as I can tell Edalize cannot do this, and this means it is not possible to use Edalize to perform remote builds (without installing it on a remote machine, which is fragught with problems).

PR #56, opened 2019-04-04
lib.io: replace TSTriple with triple_layout().

jfng:


nmigen/lib/io.py
```diff
@@ -1,27 +1,18 @@
...
-    def get_tristate(self, io):
-        return Tristate(self, io)
+class TripleLayout(Layout):
+    def __init__(self, shape):
+        fields = [
+            ("o",  shape, DIR_FANOUT),
+            ("oe", 1,     DIR_FANOUT),
+            ("i",  shape, DIR_FANIN)
+        ]
+        super().__init__(fields)
```
whitequark: I think this can be just a function `def TripleLayout`, since subclassing adds no value here.

whitequark:
This PR has been superseded by the `pin_layout` implementation.

PR #64, opened 2019-04-26
build: add ConstraintManager.

jfng:


nmigen/build/tools.py
```diff
@@ -0,0 +1,45 @@
+__all__ = ["yosys_id_escape"]
+
+
+# Implementation of the Yosys RTLIL::IdString escaping function, as defined in
+# https://github.com/YosysHQ/yosys/blob/master/backends/verilog/verilog_backend.cc#L101-L159
+def yosys_id_escape(s):
```
whitequark: This list doesn't have anything to do with Yosys itself, it is a list of (System)Verilog keywords.

nmigen/build/generic_platform.py
```diff
@@ -0,0 +1,161 @@
+from collections import defaultdict, OrderedDict
```
whitequark: This file (and its corresponding test file) should really be called something like `nmigen.build.constraint`.

nmigen/build/generic_platform.py
```diff
@@ -0,0 +1,161 @@
...
+            if r.number in self.available[r.name]:
+                raise NameError("Trying to add {!r}, but {!r} has the same name and number"
+                                .format(r, self.available[r.name][r.number]))
+            self.available[r.name][r.number] = r
+
+    def add_period(self, resource, period):
+        if not isinstance(resource, Resource):
+            raise TypeError("Object {!r} is not a Resource".format(resource))
+        if isinstance(resource.io[0], Subsignal):
+            raise ConstraintError("Cannot constrain Resource ('{}', {}) to a period of {} ns "
```
whitequark: It would be more clear to say "Cannot constrain period of resource … because it has subsignals", since there is no period that would be valid here at all.

nmigen/build/generic_platform.py
```diff
@@ -0,0 +1,161 @@
...
+__all__ = ["ConstraintError", "ConstraintManager"]
+
+
+class ConstraintError(Exception):
+    pass
+
+
+class ConstraintManager:
+    def __init__(self, resources):
+        self.available    = defaultdict(dict)
```
whitequark: First, this should be an OrderedDict too.

Second, a much nicer logic would be to have two dicts, `resources` and `requested`. In the first dict would be contained all resources ever registered with this constraint manager. In the second dict, there would be contained resources requested from this constraint manager.

The reasons are twofold. First, oMigen has a problem that if a resource is requested twice, it acts the second time as if the resource has never existed in the first place. This is very confusing. Second, with the current implementation, you can request, add, request, add, ... the same resource any amount of times.

nmigen/build/generic_platform.py
```diff
@@ -0,0 +1,161 @@
...
+        self.matched      = OrderedDict()
+        self.clocks       = OrderedDict()
+        self.tristates    = []
+        self.diffpairs    = []
+        self.ports        = []
+        self.inouts       = []
```
whitequark: Also, since none of the fields above appear to be a part of the public API, they all should be named starting with an underscore.

whitequark:
Thanks for this great PR. I've merged it manually with some changes:
  * Added XDR support.
  * Changed period constraints to frequency constraints, in Hz. There are two reasons. First, frequencies are almost always referred to in Hz; for example most clock signals in nmigen have names like `clk100`. Second, being in Hz and not MHz simplifies calculations when the same frequency value is used elsewhere, like when setting up counters.
  * Changed record creation code to use the proper interface for filling in record fields.

PR #74, opened 2019-06-02
Add tinyfpga_BX board and programmer

zignig:
Platform for tinyfpga_BX and programmer mixin.

### DIFF
--- a/nmigen/vendor/fpga/lattice_ice40.py
+++ b/nmigen/vendor/fpga/lattice_ice40.py
@@ -6,7 +6,7 @@ import tempfile
 from ...build import *
 
 
-__all__ = ["LatticeICE40Platform", "IceStormProgrammerMixin", "IceBurnProgrammerMixin","TinyProgrammerMixin"]
+__all__ = ["LatticeICE40Platform", "IceStormProgrammerMixin", "IceBurnProgrammerMixin", "TinyProgrammerMixin"]
 
 
 class LatticeICE40Platform(TemplatedPlatform):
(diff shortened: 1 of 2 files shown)


nmigen/vendor/fpga/lattice_ice40.py
```diff
@@ -132,3 +132,13 @@ def toolchain_program(self, products, name):
         with tempfile.NamedTemporaryFile(prefix="nmigen_iceburn_") as bitstream_file:
             bitstream_file.write(bitstream)
             subprocess.run([iceburn, "-evw", bitstream_file.name], check=True)
+
+
+class TinyProgrammerMixin:
+    def toolchain_program(self, products, name):
+        tinyprog= os.environ.get("ICEPROG", "tinyprog")
```
whitequark: This should say `os.environ.get("TINYPROG", ...)`

nmigen/vendor/tinyfpga_b.py
```diff
@@ -0,0 +1,34 @@
...
+
+        Resource("spiflash", 0,
+            Subsignal("cs_n", Pins("F7", dir="o")),
+            Subsignal("clk",  Pins("G7", dir="o")),
+            Subsignal("mosi", Pins("G6", dir="io")),
+            Subsignal("miso", Pins("H7", dir="io")),
+            extras=["IO_STANDARD=SB_LVCMOS33"]
+        ),
+    ]
+    prog_mode = "flash"
```
whitequark: The TinyFPGA programmer always uses the flash mode, so it doesn't need this attribute. (It's used only in `IceStormProgrammerMixin`, as you can see.)

PR #76, opened 2019-06-02
Added TinyBX gpio , fix on programmer.

zignig:


whitequark:
I don't really understand what's happening here. Why did you delete TinyFPGA B?

zignig:
Renamed it to BX to make the naming consistent.

I think my rebase got munted.

whitequark:
OK, so I've renamed it upstream in commit 37152c7.

Regarding other parts of this PR:
  1. I've added `.bin` suffix in every programmer in commit e4ebe03.
  2. I don't think there is any need to have a blinky for every single platform. The purpose of the blinky example is to demonstrate how to use the platform API in general, and I would expect anyone who intends to use nMigen to have no trouble adapting it.
  3. The "gpio" resources really should be defined as a "connector" in terms of oMigen. But there is no such feature in nMigen yet, because it needs a redesign and so it is not implemented at all yet. For now, I would say use a local patch, and once connectors are in nMigen, we can upstream that.

whitequark:
See #77.

whitequark:
And I've added GPIOs in 2ca0834d41.

PR #81, opened 2019-06-04
 fix global , and allow easy subclassing of Platform

zignig:
The global in the tinyfpga_bx board was an integer not a string. Added extend to platform to make subclassing and adding resources easier.

### DIFF
--- a/nmigen/build/plat.py
+++ b/nmigen/build/plat.py
@@ -85,6 +85,12 @@ class Platform(ConstraintManager, metaclass=ABCMeta):
 
         self._prepared   = False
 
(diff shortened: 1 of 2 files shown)


whitequark:
I fixed the typo in 45d1dc1. As for your other change, I don't think it's necessary. There is nothing wrong with using normal Python mechanisms like `super()`, and you could write it in even shorter way like this:
```python
class MyOwnPlatform(TinyFPGAPlatform):
    resources = TinyFPGAPlatform.resources + [
        Resource(...)
    ]
```

PR #84, opened 2019-06-04
compat.fhdl.specials: fix TSTriple conversion.

tdaede:
Not sure if this is correct but it does get me farther.


whitequark:
No, this is completely wrong. What are you doing exactly that you need this?

tdaede:
My old migen code uses the automatic SB_IO generation for tristates. This seemed to be what ice40_platform.py wanted for this to work, but indeed it doesn't.

whitequark:
Should be fixed in 79a37102.

PR #89, opened 2019-06-06
vendor.xilinx_s7: implement.

jfng:


nmigen/vendor/xilinx_s7.py
```diff
@@ -0,0 +1,254 @@
...
+        "{{name}}.v": r"""
+            /* {{autogenerated}} */
+            {{emit_design("verilog")}}
+        """,
+        "{{name}}.tcl": r"""
+            # {{autogenerated}}
+            create_project -force -name {{name}} -part {{platform.device}}
+            {% for file in platform.extra_files %}
+                {% if file.endswith(".v") -%}
+                    add_files {{file}}
```
whitequark: Should this also handle `.sv` files?

nmigen/vendor/xilinx_s7.py
```diff
@@ -0,0 +1,254 @@
...
+from ..hdl.ast import *
+from ..hdl.dsl import *
+from ..hdl.ir import *
+from ..build import *
+
+
+__all__ = ["XilinxS7Platform"]
+
+
+class XilinxS7Platform(TemplatedPlatform):
```
whitequark: Nit: can we call this `Xilinx7SeriesPlatform`, since that's what used in all official docs? The file would be renamed as well.

nmigen/vendor/xilinx_s7.py
```diff
@@ -0,0 +1,254 @@
...
+        * ``{{name}}_clock_utilization.rpt``:  Vivado report.
+        * ``{{name}}_route_status.rpt``: Vivado report.
+        * ``{{name}}_drc.rpt``: Vivado report.
+        * ``{{name}}_timing.rpt``: Vivado report.
+        * ``{{name}}_power.rpt``: Vivado report.
+        * ``{{name}}_route.dcp``: Vivado design checkpoint.
+        * ``{{name}}.bit``: binary bitstream.
+    """
+
+    device = abstractproperty()
```
whitequark: Is this property going to contain combined device, package and speed grade? It would be more future-proof (and to a degree more convenient) if those were split, for several reasons. First, that makes it easier to branch on device, package, etc in nMigen code portable across a family that wants to change its behavior accordingly. Second, other tools, like nextpnr, generally use separate parameters for these.

nmigen/vendor/xilinx_s7.py
```diff
@@ -0,0 +1,254 @@
...
+        * ``vivado``
+
+    Available overrides:
+        * ``script_after_read``: inserts commands after ``read_xdc`` in Tcl script.
+        * ``script_after_synth``: inserts commands after ``synth_design`` in Tcl script.
+        * ``script_after_place``: inserts commands after ``place_design`` in Tcl script.
+        * ``script_after_route``: inserts commands after ``route_design`` in Tcl script.
+        * ``script_before_bitstream``: inserts commands before ``write_bitstream`` in Tcl script.
+        * ``script_after_bitstream``: inserts commands after ``write_bitstream`` in Tcl script.
+        * ``additional_constraints``: inserts commands in XDC file.
```
whitequark: Tiny nit: I would prefer if this override was called `add_constraints`.

PR #105, opened 2019-06-16
build.plat: dedent overrides.

peteut:
`TemplatedPlatform.toolchain_prepare` should probably ` dedent` `overrides` well.


whitequark:
Hm, I guess that's harmless and useful.

PR #111, opened 2019-06-27
build.plat: fix dedent overrides.

peteut:
This fixes commit  ecf673d4a2fba24876bb9c6f0bb1768286f7b6ba for `Iterable` overrides.

### DIFF
--- a/nmigen/build/plat.py
+++ b/nmigen/build/plat.py
@@ -236,7 +236,10 @@ class TemplatedPlatform(Platform):
                 # `export VAR=` is treated on Linux.
                 return re.sub(r'^\"\"$', "", os.environ[var_env])
             elif var in kwargs:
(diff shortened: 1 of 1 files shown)


whitequark:
I don't understand what does this fix actually do.

peteut:
Some overrides, for instance [vivado_opts](https://github.com/m-labs/nmigen/blob/ecf673d4a2fba24876bb9c6f0bb1768286f7b6ba/nmigen/vendor/xilinx_7series.py#L25) expecting a list of strings, not a string. 
 ecf673d does not check for that and introduced the regression.

whitequark:
Thanks. I don't think there is any need to dedent overrides provided as a list. Can you change it so that it only dedents strings?

peteut:
Thanks and sorry for the mess.

PR #119, opened 2019-07-02
vendor.xilinx_7series: read extra .xdc files.

peteut:
I'm not quite sure if this is correct in general but I expect *.xdc* constraint extra files to be applied. This patch filters `extra_files` for `.xdc` endings and applies them using `read_xdc`.

### DIFF
--- a/nmigen/vendor/xilinx_7series.py
+++ b/nmigen/vendor/xilinx_7series.py
@@ -62,6 +62,11 @@ class Xilinx7SeriesPlatform(TemplatedPlatform):
             {% endfor %}
             add_files {{name}}.v
             read_xdc {{name}}.xdc
+            {% for file in platform.extra_files %}
+                {% if file.endswith("xdc") -%}
+                    read_xdc {{file}}
+                {% endif %}
+            {% endfor %}
             {{get_override("script_after_read")|default("# (script_after_read placeholder)")}}
             synth_design -top {{name}} -part {{platform.device}}{{platform.package}}-{{platform.speed}}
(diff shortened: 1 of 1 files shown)


whitequark:
Yes, this is reasonable. Are these commands order dependent? Should the extra `.xdc` files be read before or after the built-in ones?

peteut:
Yes, they are indeed. Additional files may refer to clocks defined in the common *.xdc* file for instance (*set_input_delay*, *set_output_delay*, ...). So I think additional *.xdc* should be read after the common one.

whitequark:
I anticipate that we will have more such loops adding extra files, so maybe we should make them nicer. Would you consider submitting another PR that abstracts this into a platform function, something like `platform.iter_extra_files(".v", ".sv")`? If not, no worries, I can do it myself.

peteut:
> I anticipate that we will have more such loops adding extra files, so maybe we should make them nicer. Would you consider submitting another PR that abstracts this into a platform function, something like `platform.iter_extra_files(".v", ".sv")`? If not, no worries, I can do it myself.

Good point, I'll look into that.

peteut:
@whitequark like 4c61207e717e14dccb43afd9ca4d34be1f580b7e? 
If so I'll update all platforms to employ the `iter_extra_files` method and create a PR.

whitequark:
@peteut Pretty much! Just a couple of nits.

peteut:
> @peteut Pretty much! Just a couple of nits.

@whitequark thanks for reviewing it, I'll fix it tonight.

PR #163, opened 2019-08-02
Specific error message for missing module return from 'elaborate'.

ret:
As a beginner, I made this mistake a few times, when I was commenting things in and out from my 'elaborate' functions. When I forgot to return a module from 'elaborate', I got the error 'AttributeError: Object 'None' cannot be elaborated.', which threw me off initially. Perhaps this can help other beginners to locate the problem quickly.

### DIFF
--- a/nmigen/hdl/ir.py
+++ b/nmigen/hdl/ir.py
@@ -67,7 +67,10 @@ class Fragment:
                     stacklevel=2)
                 obj = obj.elaborate(platform)
             else:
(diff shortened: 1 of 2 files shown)


ret:
Squashed the two changes into one '0f2652096eb7269af36f82b7be69d313702d45b5'.

ret:
I'll redo this properly. Will resend a new pull request shortly.

whitequark:
Don't worry, I'll just apply this myself, it would be easier that way.

PR #176, opened 2019-08-20
vendor.lattice_ecp5: add Diamond support (WIP)

### LINKED ISSUE #114: Support building ECP5 bitstreams with Diamond
There are still some places where Diamond produces superior synthesis and PNR results than Yosys+nextpnr, so it's important to be able to compare these. I've added all Diamond IOBs to Yosys so the IOB generation code shouldn't be adjusted at all; only scripts.

whitequark:
This is a work in progress. It has the following issues:
  - [x] Trellis uses `INIT` parameter for `LUT4`, and Diamond uses `init` (note the case difference)
  - [x] SVF files are not generated
  - [x] BIT and SVF files are not placed in the root of `build/`
  - [x] No overrides are available (probably need some for bitgen?)


whitequark:
This is almost ready, other than for the LUT INIT issue. Actually, I don't think we need the LUTs there in the first place...

whitequark:
@mithro By the way, this is an example of why I am reluctant to add many combinatorial toolchain configurations: just look at the amount of code added here.

PR #178, opened 2019-08-21
vendor.intel: add Quartus support

whitequark:
This is a work in progress. It has the following issues:
  - [x] SDR IO is not supported
  - [x] DDR IO is not suppored
  - [x] Differential IO is not supported
  - [x] Startup clock gating is not performed


ZirconiumX:
Regarding "startup clock gating" here, do you mean power-on reset? [This](https://github.com/intel/de10-nano-hardware/blob/master/ip/power_on_reset/power_on_reset.v#L25) appears to be the standard way of performing power-on reset, by instantiating a shift chain to hold a reset line.

PR #179, opened 2019-08-21
Xilinx 7 series STARTUPE3->STARTUPE2

dlharmon:
STARTUPE3 is for Ultrascale and it fails to build for 7 series with this in there


whitequark:
Thanks. Indeed I have misplaced some S7 docs with US.

PR #181, opened 2019-08-22
add support for real (float) parameters to Instance

dlharmon:
required for Xilinx MMCME2_BASE, etc


whitequark:
This is actually the wrong format, and it will truncate precision:
```
>>> "{:f}".format(1.23456789)
'1.234568'
>>> "{!r}".format(1.23456789)
'1.23456789'
```

whitequark:
Fixed in 47bad3d.

PR #212, opened 2019-09-20
Xilinx specific MultiReg

dlharmon:
Vivado was inferring an SRL16 from a MultiReg in some cases. This simply adds an `ASYNC_REG="TRUE"` attribute to the registers.

This code is identical for both 7 series and Spartan which I don't like. I see two possibilities: 

Add something like async_reg_attrs to the platform file and if present apply it to the regs in the MultiReg. It could also be applied to ResetSynchronizer, etc.

A new file, vendor/xilinx_common.py containing code common to all supported Xilinx parts.

I'm happy to do either.

### DIFF
--- a/nmigen/vendor/xilinx_7series.py
+++ b/nmigen/vendor/xilinx_7series.py
@@ -360,3 +360,11 @@ class Xilinx7SeriesPlatform(TemplatedPlatform):
                 io_IO=p_port[bit], io_IOB=n_port[bit]
             )
         return m
+
+    def get_multi_reg(self, multireg):
+        m = Module()
+        for i, o in zip((multireg.i, *multireg._regs), multireg._regs):
+            o.attrs["ASYNC_REG"] = "TRUE"
+            m.d[multireg._o_domain] += o.eq(i)
+        m.d.comb += multireg.o.eq(multireg._regs[-1])
+        return m
--- a/nmigen/vendor/xilinx_spartan_3_6.py
+++ b/nmigen/vendor/xilinx_spartan_3_6.py
@@ -411,6 +411,14 @@ class XilinxSpartan3Or6Platform(TemplatedPlatform):
             )
         return m
 
+    def get_multi_reg(self, multireg):
+        m = Module()
+        for i, o in zip((multireg.i, *multireg._regs), multireg._regs):
+            o.attrs["ASYNC_REG"] = "TRUE"
+            m.d[multireg._o_domain] += o.eq(i)
+        m.d.comb += multireg.o.eq(multireg._regs[-1])
+        return m
+
 
 XilinxSpartan3APlatform = XilinxSpartan3Or6Platform
 XilinxSpartan6Platform = XilinxSpartan3Or6Platform


whitequark:
> This code is identical for both 7 series and Spartan which I don't like.

This is actually just fine. Duplicating code is cheap, untangling code when you e.g. want to refactor Vivado and have no access to ISE is very expensive, in terms of time and bugs.

whitequark:
Actually, have you checked that this is relevant to ISE? Your commit message only mentions Vivado.

dlharmon:
I have not checked on ISE as I don't currently have it installed. The language templates in ISE call for the same thing and I have used it in the past on Spartan 6 (Verilog). I'm happy to remove it from the Spartan parts if you prefer.

whitequark:
> I have not checked on ISE as I don't currently have it installed.

I have ISE, but do not know Xilinx tooling well. I can check it if you tell me what to look for to determine whether everything works correctly (ideally an nMigen design I can just run.)

Incidentally, this is exactly why I would like to *not* have `xilinx_common.py`. It is much better to have duplicated code that is tested on each platform it is added to, than factored out code which is only tested on a subset of platforms when it is added or chnaged.

dlharmon:
ISE is installing. I'll check it. MultiReg with n=4 was getting turned into an SRL16 with a delay of 3 and a single FDRE.

Your approach to keeping the tools separate makes sense. I've worked on projects in the past that avoided code duplication at all costs and that was stuck in my head.

whitequark:
You might find this [essay](https://programmingisterrible.com/post/139222674273/write-code-that-is-easy-to-delete-not-easy-to) on the topic insightful.

dlharmon:
I've confirmed the same behavior on Spartan6/ISE.

Test code:
```python3
from nmigen import *
from nmigen.lib.cdc import MultiReg
from nmigen_boards.mercury import *

class Blinky(Elaboratable):
    def elaborate(self, platform):
        clk50   = platform.request("clk50")
        user_led = platform.request("bussw_oe", 0)
        btn = platform.request("user_btn", 0)
        counter  = Signal(20)

        m = Module()
        m.domains.sync = ClockDomain()
        m.d.comb += ClockSignal().eq(clk50.i)
        btn_sync = Signal()
        m.submodules.mreg = MultiReg(i=btn, o=btn_sync, o_domain="sync", n=2, reset=1)
        m.d.sync += counter.eq(Mux(btn_sync, 0, counter + 1))
        m.d.comb += user_led.o.eq(counter[-1])

        return m

if __name__ == "__main__":
    platform = MercuryPlatform()
    platform.build(Blinky(), do_program=False)
```

With patch, no SRLs are inferred. Without, the MultiReg becomes an SRL16 + FDRE.

whitequark:
Would you like to take a look at `ResetSynchronizer` too, given that 0.1 is going to be released soon?

dlharmon:
`ResetSynchronizer` has an asynchronous reset on the flip flops which prevents SRL16s from being inferred. I tested on ISE, but it should be the case on Vivado as well as there is no way to reset bits internal to an SRL16. ISE does complain about not being able to use an SRL when n > 2. Giving `ResetSynchronizer` the same treatment as `MultiReg` would silence that, but it's probably not worth having the additional code.

Test code:
```python3
from nmigen import *
from nmigen.lib.cdc import ResetSynchronizer
from nmigen_boards.mercury import *

class Blinky(Elaboratable):
    def elaborate(self, platform):
        clk50   = platform.request("clk50")
        user_led = platform.request("bussw_oe", 0)
        btn = platform.request("user_btn", 0)
        counter  = Signal(20)

        m = Module()
        m.domains.sync = ClockDomain()
        m.d.comb += ClockSignal().eq(clk50.i)
        btn_sync = Signal()
        m.submodules.rsync = ResetSynchronizer(arst=btn, n=4)
        m.d.sync += counter.eq(counter + 1)
        m.d.comb += user_led.o.eq(counter[-1])

        return m

if __name__ == "__main__":
    platform = MercuryPlatform()
    platform.build(Blinky(), do_program=False)

```

whitequark:
Thanks for investigating this. What about false paths for both MultiReg and ResetSynchronizer?

dlharmon:
I'm happy to work on that and it is something I would benefit from. It's a bit bigger problem and may be a week or two before I get time. I don't use ISE on new designs, so would be tempted to just do it for Vivado and maybe ECP5/nextpnr.

whitequark:
I think that's a reasonable approach.

dlharmon:
Looking into timing constraints / false paths I found that `ResetSynchronizer` would benefit from the `ASYNC_REG="TRUE"` attribute as well on both Vivado and ISE. It will cause the registers in the chain to be located as close as possible to each other which is beneficial for minimizing metastabilty. I can create a pull request for ISE if desired. Vivado will be done as part of clock domain crossing constraints pull request which is coming shortly.

whitequark:
> I can create a pull request for ISE if desired.

Please do.

PR #221, opened 2019-09-21
vendor.altera: use buffer primitives

ZirconiumX:
This adds SDR support for Altera chips, or at least enough flops and buffers for it.

### DIFF
--- a/nmigen/vendor/altera.py
+++ b/nmigen/vendor/altera.py
@@ -142,134 +142,186 @@ class AlteraPlatform(TemplatedPlatform):
             i_prn=1,
             o_q=dest
         )
-    
+  
     # Despite the altiobuf manual saying ENABLE_BUS_HOLD is optional, Quartus requires it to be specified.
 
     def get_input(self, pin, port, attrs, invert):
         self._check_feature("single-ended input", pin, attrs,
-                            valid_xdrs=(0,1), valid_attrs=True)
+                            valid_xdrs=(0,1,2), valid_attrs=True)
 
         m = Module()
 
         ff_i = Signal(pin.width)
 
-        if pin.xdr == 1:
+        if pin.xdr >= 1:
+            pin.i.attrs["useioff"] = "1"
+
+        for bit in range(pin.width):
+            if pin.xdr <= 1:
+                clk = pin.i_clk if pin.xdr != 0 else None
+
+                self._add_ff(m, pin.xdr, self._invert_if(invert, ff_i[bit]), pin.i[bit], clk, "i")
+
+                m.submodules["{}_buf_{}".format(pin.name, bit)] = Instance("altiobuf_in",
+                    p_NUMBER_OF_CHANNELS=1,
+                    p_ENABLE_BUS_HOLD="FALSE",
+                    i_datain=port[bit],
+                    o_dataout=ff_i[bit]
+                )
+            else:
+                m.submodules["{}_buf_{}".format(pin.name, bit)] = Instance("altddio_in",
+                    p_width=1,
+                    i_datain=port[bit],
+                    i_inclock=pin.i_clk,
+                    o_dataout_h=self._invert_if(pin.i0[bit]),
+                    o_dataout_l=self._invert_if(pin.i1[bit]),
+                )
+
+        return m
+
+    def get_output(self, pin, port, attrs, invert):
+        self._check_feature("single-ended output", pin, attrs,
+                            valid_xdrs=(0,1,2), valid_attrs=True)
+
+        m = Module()
+
+        ff_o = Signal(pin.width)
+
+        if pin.xdr >= 1:
+            pin.o.attrs["useioff"] = "1"
+
+        for bit in range(pin.width):
+            if pin.xdr <= 1:
+                clk = pin.o_clk if pin.xdr != 0 else None
+
+                self._add_ff(m, pin.xdr, self._invert_if(invert, pin.o[bit]), ff_o[bit], clk, "o")
+
+                m.submodules["{}_buf_{}".format(pin.name, bit)] = Instance("altiobuf_out",
+                    p_NUMBER_OF_CHANNELS=1,
+                    p_ENABLE_BUS_HOLD="FALSE", 
+                    i_datain=ff_o[bit],
+                    o_dataout=port[bit]
+                )
+            else:
+                m.submodules["{}_buf_{}".format(pin.name, bit)] = Instance("altddio_out",
+                    p_width=1,
+                    i_datain_h=self._invert_if(invert, pin.o0[bit]),
+                    i_datain_l=self._invert_if(invert, pin.o1[bit]),
+                    i_outclock=pin.o_clk,
+                    o_dataout=port[bit]
+                )
+                
+        return m
+
+    def get_tristate(self, pin, port, attrs, invert):
+        self._check_feature("single-ended tristate", pin, attrs,
+                            valid_xdrs=(0,1,2), valid_attrs=True)
+
+        m = Module()
+
+        ff_o = Signal(pin.width)
+        ff_oe = Signal(pin.width)
+
+        if pin.xdr >= 1:
+            pin.o.attrs["useioff"] = "1"
+            pin.oe.attrs["useioff"] = "1"
+
+        for bit in range(pin.width):
+            if pin.xdr <= 1:
+                clk = pin.o_clk if pin.xdr != 0 else None
+
+                self._add_ff(m, pin.xdr, self._invert_if(invert, pin.o[bit]), ff_o[bit], clk, "o")
+                self._add_ff(m, pin.xdr, pin.oe[bit], ff_oe[bit], clk, "oe")
+
+                m.submodules["{}_buf_{}".format(pin.name, bit)] = Instance("altiobuf_out",
+                    p_NUMBER_OF_CHANNELS=1,
+                    p_ENABLE_BUS_HOLD="FALSE", 
+                    p_USE_OE="TRUE",
+                    i_datain=ff_o[bit],
+                    i_oe=ff_oe[bit],
+                    o_dataout=port[bit]
+                )
+            else:
+                m.submodules["{}_buf_{}".format(pin.name, bit)] = Instance("altddio_out",
+                    p_width=1,
+                    p_oe_reg="REGISTERED",
+                    i_datain_h=self._invert_if(invert, pin.o0[bit]),
+                    i_datain_l=self._invert_if(invert, pin.o1[bit]),
+                    i_outclock=pin.o_clk,
+                    i_oe=pin.oe[bit],
+                    o_dataout=port[bit]
+                )
+
+        return m
+
+    def get_input_output(self, pin, port, attrs, invert):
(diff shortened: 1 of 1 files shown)


nmigen/vendor/altera.py
```diff
@@ -121,45 +121,140 @@ def _invert_if(invert, value):
...
+            elif pin.xdr == 1:
+                m.submodules["{}_dff_i_{}".format(pin.name, bit)] += Instance("dff",
+                    i_d=self._invert_if(invert, ff_i[bit]),
+                    i_clk=pin.i_clk,
+                    i_clrn=1,
+                    i_prn=1,
+                    o_q=pin.i[bit]
+                )
+
         return m
```
peteut: According [AN307](https://www.intel.com/content/dam/www/programmable/us/en/pdfs/literature/an/an307.pdf), table 56 you should probably set the attribute `FAST_INPUT_REGISTER` to use the reg within the IOB.

nmigen/vendor/altera.py
```diff
@@ -121,45 +121,140 @@ def _invert_if(invert, value):
...
+                    i_clrn=1,
+                    i_prn=1,
+                    o_q=ff_o[bit]
+                )
+
+            m.submodules["{}_buf_{}".format(pin.name, bit)] = Instance("alt_outbuf",
+                i_i=ff_o[bit],
+                o_o=port[bit]
+            )
+
```
peteut: According [AN307](https://www.intel.com/content/dam/www/programmable/us/en/pdfs/literature/an/an307.pdf), table 56 you should probably set the attribute `FAST_OUTPUT_REGISTER` to use the reg within the IOB.

ZirconiumX:
So, `FAST_{INPUT,OUTPUT}_REGISTER` isn't an attribute, but a TCL command, so `top.qsf` would need to be modified for that. I'm not sure best how to do that.

whitequark:
@ZirconiumX Looks like you can use the `altera_attribute` attribute in Verilog, something like `(* altera_attribute = "-name FAST_INPUT_REGISTER ON" *)`, see [here](https://www.intel.com/content/www/us/en/programmable/quartushelp/17.0/hdl/vlog/vlog_file_dir_attribute.htm). But also, it looks like in this case you can use `(* useioff *)` attribute instead, see [here](https://www.intel.com/content/www/us/en/programmable/quartushelp/17.0/hdl/vlog/vlog_file_dir.htm).

Note that you still need to verify whether those will work on `alt_*buf` instances.

whitequark:
See also [this post](http://billauer.co.il/blog/2017/04/altera-intel-fpga-io-ff-packing/).

nmigen/vendor/altera.py
```diff
@@ -137,6 +137,7 @@ def get_input(self, pin, port, attrs, invert):
                 m.d.comb += pin.i[bit].eq(self._invert_if(invert, ff_i[bit]))
             elif pin.xdr == 1:
                 m.submodules["{}_dff_i_{}".format(pin.name, bit)] += Instance("dff",
+                    a_useioff=True,
```
whitequark: Does this actually work?

ZirconiumX:
I fixed the mangled branch history, sorry.

nmigen/vendor/altera.py
ZirconiumX: I have no way of testing SDR, let alone this specifically, and I don't have an oscilloscope to check that SDR output is sane, so probably not.

whitequark: Does this synthesize without complaints from Quartus? What if you mistype the attribute on purpose, does it complain then?

ZirconiumX: > Does this synthesize without complaints from Quartus?
Quartus appears to recognise but ignore the attribute; I had to consult the manual to find out that what Quartus expects is not e.g.
```verilog
module top(input clk50_0__io, input i, output o);
wire ff_o;
(* useioff = 1 *) dff foo (.d(i), .clk(clk50_0__io), .clrn(1'b1), .prn(1'b1), .q(ff_o));
alt_outbuf bar (.i(ff_o), .o(o));
endmodule
```
but
```verilog
module top(input clk50_0__io, (* useioff = 1 *) input i, output o);
wire ff_o;
dff foo (.d(i), .clk(clk50_0__io), .clrn(1'b1), .prn(1'b1), .q(ff_o));
alt_outbuf bar (.i(ff_o), .o(o));
endmodule
```
which gives you
```
Info (176235): Finished register packing
    Extra Info (176218): Packed 1 registers into blocks of type I/O input buffer
```
in the compiler output.

I'm not sure how to express that the top-level arguments of a module/fragment need attributes though.

> What if you mistype the attribute on purpose, does it complain then?
`Warning (10335): Unrecognized synthesis attribute "useiofff" at top.v(5) File: //wsl$/Ubuntu-18.04/home/lofty/de_10/build/top.v Line: 5`

whitequark: Try `pin.i.attrs["useioff"] = "1".

ZirconiumX: Like 15ac723?

whitequark:
Looks good. The next steps would be to add DDR support and then factor out common code so that it's not quite as repetitive. (I think DDR suppor would influence the way you can refactor this.)

ZirconiumX:
Should I put that in a separate PR, or just continue this?

whitequark:
Continue this.

ZirconiumX:
I've been working from [UG83105](https://www.intel.com/content/dam/www/programmable/us/en/pdfs/literature/ug/ug_low_level.pdf), which also mentions the differential I/O primitives. However, I still need to hunt for what a useful DDR primitive would be, given that using IP blocks seems troublesome.

whitequark:
I think you might be able to generate IP cores with `qsys-script` and `qsys-generate` as described [here](https://www.intel.com/content/dam/www/programmable/us/en/pdfs/literature/ug/ug_intro_to_megafunctions.pdf).

ZirconiumX:
I'm unsure if this should advertise SDR support or not at the moment. Thoughts?

whitequark:
It should. I won't merge the complete backend (i.e. the `vendor.altera` branch, this PR is fine) until SDR and DDR are verified working, anyway.

ZirconiumX:
While writing f24137b a question occurred to me: what's the correct clock to use for an output-enable?

whitequark:
The output clock.

ZirconiumX:
I realised I added a bug in 4c7eb1b by using the clock fields without checking for their existence first. The solution in 37bb61c feels a bit hacky but I'm not sure of anything better.

whitequark:
Have you looked at how the other vendor packages solve this? E.g. Xilinx? I would expect that a similar code structure would work for Altera.

ZirconiumX:
I have, but I found them very confusing and difficult to understand. You're welcome to refactor my code to fit in better though.

whitequark:
Sure, but you'll need to add DDR support first, even if it doesn't work.

ZirconiumX:
Writing ba3dc3c felt like wading through mud; did Altera *need* to make the cells this difficult to use? I've not done tristate/bidirectional stuff yet because I'm not sure I can summon the motivation for it; it's just painful to work on.

ZirconiumX:
I've done some more research over the `altlvds` cell, and I'm not convinced its semantics match up with nMigen's; the I/O cells for nMigen seem to be implicitly infallible without initialisation, but `altlvds` is fallible (PLLs may desynchronise causing data corruption) and requires initialisation (for PLL lock).

whitequark:
`altlvds` also has sparse availability. It should not be necessary to use a PLL to make a DDR receiver on any FPGA family, Altera or not...

ZirconiumX:
Yeah it's not great, but I don't know of any better solutions.

whitequark:
Well I'm not actually convinced yet that `altddio` doesn't support differential receive pins. What does Quartus say about your combination of `altddio` and `altiobuf`?

ZirconiumX:
It doesn't like it. Given:
```verilog
module top(input clk, p_port, n_port, output i0, i1);

wire p_posedge, p_negedge, n_posedge, n_negedge;

altddio_in #(.width(1)) foo_ddr_p_0 (.datain(p_port), .inclock(clk), .dataout_h(p_posedge), .dataout_l(p_negedge));
altddio_in #(.width(1)) foo_ddr_n_0 (.datain(n_port), .inclock(clk), .dataout_h(n_posedge), .dataout_l(n_negedge));
altiobuf_in #(.NUMBER_OF_CHANNELS(1), .ENABLE_BUS_HOLD("FALSE"), .USE_DIFFERENTIAL_MODE("TRUE")) foo_buf_p_0 (.datain(p_posedge), .datain_b(n_posedge), .dataout(i0));
altiobuf_in #(.NUMBER_OF_CHANNELS(1), .ENABLE_BUS_HOLD("FALSE"), .USE_DIFFERENTIAL_MODE("TRUE")) foo_buf_n_0 (.datain(p_negedge), .datain_b(n_negedge), .dataout(i1));

endmodule
```

I get
```
Error (17044): Illegal connection on I/O input buffer primitive altiobuf_in:foo_buf_n_0|iobuf_in_l9g:auto_generated|ibuf
a[0]. Source I/O pin altddio_in:foo_ddr_p_0|ddio_in_d0b:auto_generated|ddio_ina[0] drives out to destinations other than
 the specified I/O input buffer primitive. Modify your design so the specified source I/O pin drives only the specified
I/O input buffer primitive. File: //wsl$/Ubuntu-18.04/home/lofty/de_10/build/db/iobuf_in_l9g.tdf Line: 33
```

whitequark:
OK. Looks like it is indeed not supported. So you should set `xdrs=(0,1)` for differential pins and `xdrs=(0,1,2)` for single-ended ones.

ZirconiumX:
```verilog
module top(clk50_0__io, p_port, n_port, i0, i1);
input clk50_0__io;
(* altera_attribute = "-name IO_STANDARD \"LVDS\" " *) input p_port, n_port;
output i0, i1;

wire buf_out;

altiobuf_in #(.NUMBER_OF_CHANNELS(1), .ENABLE_BUS_HOLD("FALSE"), .USE_DIFFERENTIAL_MODE("TRUE")) foo_buf_p_0 (.datain(p_port), .datain_b(n_port), .dataout(buf_out));
altddio_in #(.width(1)) foo_ddr_p_0 (.datain(buf_out), .inclock(clk50_0__io), .dataout_h(i0), .dataout_l(i1));

endmodule
```

This works okay. Guess the altddio documentation is wrong here.

ZirconiumX:
Worth noting is that Quartus is extremely picky about the I/O standard here. Without the `altera_attribute` there, this code fails to compile.

whitequark:
That's normal. It should already be applied by nMigen elsewhere in platform code.

ZirconiumX:
Presumably then I need to implement DDR tristate/bidirectional I/O for completeness?

ZirconiumX:
Okay, I *believe* this is feature-complete. The resulting Verilog needs to be checked and tested though.

whitequark:
Looks good at a quick glance. Thank you for the contribution. I will check this on real hardware at some point soon, and then merge while adjusting anything that needs changing.

PR #234, opened 2019-09-24
make the set_max_delay value a string

dlharmon:
If `max_input_delay` was set to 5.3, the verilog was `(* \nmigen.vivado.max_delay  = 32'd5 *)`

With this change, the verilog is `(* \nmigen.vivado.max_delay  = "5.3" *)`


whitequark:
Actually, there's one more change I want to do to it. Since everywhere else in nMigen frequencies are specified in hertz, then max delay should be specified in seconds.

PR #241, opened 2019-09-29
rpc: add public Records as module ports.

jfng:
Use case: modules exposing streams or bus interfaces as ports.


whitequark:
Thank you!

PR #246, opened 2019-10-09
vendor.intel: reset synchronisers

ZirconiumX:
I couldn't actually find any documentation on this, so I'm working based on the simulation model for this, and hoping for the best.

### DIFF
--- a/nmigen/vendor/intel.py
+++ b/nmigen/vendor/intel.py
@@ -371,3 +371,36 @@ class IntelPlatform(TemplatedPlatform):
             i_oe=self._get_oereg(m, pin),
         )
         return m
+
+    def get_ff_sync(self, ff_sync):
+        assert ff_sync._stages >= 2, "the synchronizer must be at least 2 stages long"
+
+        m = Module()
+        m.submodules += Instance("altera_std_synchronizer",
+            p_depth=ff_sync._stages, 
+            i_clk=ClockSignal(ff_sync._o_domain),
+            i_reset_n=1,
+            i_din=ff_sync.i,
+            o_dout=ff_sync.o,
+        )
+        return m
+
+    def get_reset_sync(self, reset_sync):
+        assert reset_sync._stages >= 2, "the synchronizer must be at least 2 stages long"
+
+        m = Module()
+        reset = Signal()
+        reset_clk = ClockSignal(reset_sync._domain)
+        m.submodules += Instance("altera_std_synchronizer",
+            p_depth=reset_sync._stages,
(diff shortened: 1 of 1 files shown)


nmigen/vendor/intel.py
```diff
@@ -371,3 +371,36 @@ def get_diff_input_output(self, pin, p_port, n_port, attrs, invert):
             i_oe=self._get_oereg(m, pin),
         )
         return m
+
+    def get_ff_sync(self, ff_sync):
+        assert ff_sync._stages >= 2, "the synchronizer must be at least 2 stages long"
```
whitequark: This isn't the place to add this assert (maybe FFSynchronizer itself is, but not in this PR).

nmigen/vendor/intel.py
```diff
@@ -371,3 +371,36 @@ def get_diff_input_output(self, pin, p_port, n_port, attrs, invert):
...
+        m.submodules += Instance("altera_std_synchronizer",
+            p_depth=reset_sync._stages,
+            i_clk=reset_clk,
+            i_reset_n=~reset_sync.arst,
+            i_din=1,
+            o_dout=reset,
+        )
+        m.d.comb += [
+            ClockSignal("reset_sync").eq(reset_clk),
+            ResetSignal("reset_sync").eq(reset_sync.arst),
```
whitequark: I don't understand what this logic is doing.

ZirconiumX: I'm cross-referencing Xilinx here which also uses a flop chain. It *is* a bit confusing though.

whitequark: The Xilinx code adds a local clock domain:

    m.domains += ClockDomain("reset_sync", async_reset=True, local=True)

That's not necessary here, you can ditch everything related to this domain.

ZirconiumX: Ah, I see, thank you.

nmigen/vendor/intel.py
```diff
@@ -371,3 +371,28 @@ def get_diff_input_output(self, pin, p_port, n_port, attrs, invert):
...
+            i_reset_n=1,
+            i_din=ff_sync.i,
+            o_dout=ff_sync.o,
+        )
+        return m
+
+    def get_reset_sync(self, reset_sync):
+        m = Module()
+        reset = Signal()
+        reset_clk = ClockSignal(reset_sync._domain)
```
whitequark: I think you can just use `ClockSignal` and `ResetSignal` directly in instance ports.

PR #266, opened 2019-11-08
Record: force name to be given by name for `__init__()`

Fatsie:
This is to make it easier to support the `like()` class method in subclasses of Record. If in the subclass the second argument to `__init__()` is not name one currently has to take into account that the second argument can actually be the name if the first argument is of type `Layout`.

### DIFF
--- a/nmigen/hdl/rec.py
+++ b/nmigen/hdl/rec.py
@@ -107,9 +107,9 @@ class Record(Value):
                 fields[field_name] = Signal.like(field, name=concat(new_name, field_name),
                                                  src_loc_at=1 + src_loc_at)
 
-        return cls(other.layout, new_name, fields=fields, src_loc_at=1)
+        return cls(other.layout, name=new_name, fields=fields, src_loc_at=1)
 
-    def __init__(self, layout, name=None, *, fields=None, src_loc_at=0):
+    def __init__(self, layout, *, name=None, fields=None, src_loc_at=0):
         if name is None:
             name = tracer.get_var_name(depth=2 + src_loc_at, default=None)
 


whitequark:
I think the change is good (everywhere else `name` is keyword-only already), but I have no idea what your PR description is supposed to mean.

Fatsie:
Updated comment, added:

For example if you have:

```python
class JTAG(Record):
    def __init__(master=False, trst=False, *, name=None, src_loc_at=0):
```

If one uses the .like() function on a JTAG object it will pass name in trst
and not in name without this patch.

Fatsie:
After some reflection I think some more documentation or changes need to be done to handle calling .like() for subclasses of Record.
For the same JTAG class, if one does:
```python
jtag = JTAG()
jtag_like = Record.like(jtag)
```
jtag_like is of type Record and not JTAG. Is this wanted behavior ?

whitequark:
I've merged the PR but with a completely different rationale, see the updated commit message.

> jtag_like is of type Record and not JTAG. Is this wanted behavior ?

Yes. There is no general way to handle this such that `like` returns an instance of the subclass. The generic `Record.like` should be changed to use `Record(...)` instead of `cls(...)` and any subclasses that want `like` to return a subclass instance should reimplement it themselves.

Fatsie:
> Yes. There is no general way to handle this such that like returns an instance of the subclass. The generic Record.like should be changed to use Record(...) instead of cls(...) and any subclasses that want like to return a subclass instance should reimplement it themselves.

I do think there is a way to do it if you force that subclasses make their `.like()` parameters compatible with the parent:
```python
class Record(Value):
    @classmethod
    def like(cls, other, *, name=None, name_suffix=None, src_loc_at=0):
        assert isinstance(other, cls)
        if cls != other.__class__:
            return other.__class__.like(other, name=name, name_suffix=name_suffix, src_loc_at=src_loc_at)
```

Reason I don't like that calling `Record.like(jtag)` giving resulting in a Record object is interaction with overloaded methods. If JTAG overloads Record methods like for example `connect()` that overloaded behavior will not be seen on the object coming from `Record.like()`.
The question is actually what the use case is for Record.like() and what can and can't be done with the result of it.
Should `.like()` actually need to be a class method ? Can't it just be a normal method that then can be overloaded by the JTAG class ?

whitequark:
> I do think there is a way to do it if you force that subclasses make their `.like()` parameters compatible with the parent:

The branch in your example will never fire, unless I'm misreading it.

> If JTAG overloads Record methods like for example `connect()` that overloaded behavior will not be seen on the object coming from `Record.like()`.

You should in general not overload `connect()`, definitely not if you violate LSP when doing it. In general, the use of `Record.connect` is ill-advised and almost always a dedicated connector module or a dedicated method with record-specific behavior is a better idea.

> Should `.like()` actually need to be a class method ? Can't it just be a normal method that then can be overloaded by the JTAG class ?

You can override class methods just fine.

whitequark:
Maybe we can solve the `Record.like` problem by adding a method that returns the arguments to `__init__` for that subclass. So for `Record` itself it would return `{"layout": self.layout}`, for your subclass it would return `{"master": ..., "trst": ...}`.

Fatsie:
> The branch in your example will never fire, unless I'm misreading it.

No, if you call `Record.like(jtag)`, `cls` will be `Record` and `other.__class__` will be JTAG. Or am I missing something ?

> Maybe we can solve the `Record.like` problem by adding a method that returns the arguments to `__init__` for that subclass. So for Record itself it would return {"layout": self.layout}, for your subclass it would return `{"master": ..., "trst": ...}`.

and then use `other.__class__(**kwargs)` ? Could be solution.

whitequark:
Ah, right.

Fatsie:
But maybe nothing has to be changed, just a comment in the code
```python
class Record(Value):
    @classmethod
    def like(cls, other, *, name=None, name_suffix=None, src_loc_at=0):
        """Return fresh object with same fields as other object

        Use Record.like(other) if one wants a Record object with same fields as other.
        Use other.__class__.like(other) if one wants an object of same class as other
        and with same fields."""
```

whitequark:
I like that. I'll make sure to mention it when I add some docs for `Record`, which I was going to do soon.

PR #267, opened 2019-11-09
cli: add option to build design

tpwrules:
Adds a subcommand to build a design on a platform, with the option to program the platform too. 

One potential wrinkle: if the argument parsing fails, the design is destroyed unceremoniously and there is an ugly `UnusedElaboratable` warning. Maybe `_Elaboratable__silence` should be set to True on the design? This would require the `main` function to be aware of that argument.


tpwrules:
Turns out to not fit well with the vision.

PR #268, opened 2019-11-12
class Memory: allow to set default `simulate` value for all generated memory blocks.

Fatsie:
I have nmigen code mixed with other RTL language code so am using external simulator. With this option I can now turn off the memory preparation for pysim for all generated blocks.
```python
from nmigen import *

Memory.simulate = False
```


whitequark:
I'm rewriting the simulator right now so that it would not need the (expensive) memory lowering.

PR #272, opened 2019-11-18
vendor.xilinx_*: Set IOB attribute on cells instead of nets.

### LINKED ISSUE #263: On Xilinx platforms, IOB should be set on the cell, not the net
According to @jordens: https://github.com/m-labs/artiq/pull/1382#issuecomment-547823173

I think the code in question predates the possibility to add attributes to cells.

jfng:
Fixes #263.


whitequark:
Is it not sufficient to use `a_IOB=1` on the `Instance` for some reason?

whitequark:
Thanks!

PR #277, opened 2019-11-30
vendor.intel: silence meaningless warnings

ZirconiumX:
Quartus complains a lot while building Yosys-generated Verilog:

- it fails some subjective lints (warnings 10270, 10763, 10935) that don't affect the semantics (modulo Yosys bugs).
- Quartus doesn't recognise certain attributes like `src` (warning 10335) generated by nMigen.
- Quartus will use its internal IP library whenever you instantiate a megafunction, and warn about it not being within the settings file (warning 12125).
- Quartus complains whenever you set `NUM_PARALLEL_PROCESSORS` to something which isn't all cores (warning 18236).
- Quartus Lite will also complain about unimportant features only being available when you buy Quartus (warning 292013).

These can be silenced through the creation of a `.srf` file (which in this case suppresses all message text with a given ID in `quartus_suppressed_warnings`).

Of these, all but 10335 are effectively harmless. Silencing 10335 has the potential to obscure a misspelt user attribute, so I could remove it if you'd prefer.

### DIFF
--- a/nmigen/vendor/intel.py
+++ b/nmigen/vendor/intel.py
@@ -38,6 +38,16 @@ class IntelPlatform(TemplatedPlatform):
     speed   = abstractproperty()
     suffix  = ""
 
+    quartus_suppressed_warnings = [
+        10270,  # Incomplete Verilog case statement has no default case item
+        10335,  # Unrecognized synthesis attribute
+        10763,  # Verilog case statement has overlapping case item expressions with non-constant or don't care bits
+        10935,  # Verilog casex/casez overlaps with a previous casex/vasez item expression
+        12125,  # Using design file which is not specified as a design file for the current project, but contains definitions used in project
+        18236,  # Number of processors not specified in QSF
+        292013, # Feature is only available with a valid subscription license
+    ]
+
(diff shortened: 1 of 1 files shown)


whitequark:
> * Quartus doesn't recognise certain attributes like `src` (warning 10335) generated by nMigen.

Hang on. Aren't these stripped? #220 should have fixed that.

ZirconiumX:
Indeed they are; my nMigen was a little out of date.

ZirconiumX:
```
Info (10264): Verilog HDL Case Statement information at top.v(669): all case item expressions in this case statement are
 onehot File: //wsl$/Ubuntu-18.04/home/lofty/runelord/gateware/runelord/build/top.v Line: 669
```
This should probably be silenced.

whitequark:
Sure. Is there any way to silence those Verilog warnings only for nMigen-emitted code?

ZirconiumX:
One of the fields seems to be a keyword match. We can include the nMigen-emitted filenames in that.

whitequark:
Sounds OK.

PR #279, opened 2019-12-03
hdl.ast: Fix width for unary minus operator on signed argument.

mwkmwkmwk:
To properly represent a negation of a signed X-bit quantity we may, in
general, need a signed (X+1)-bit signal — for example, negation of
3-bit -4 is 4, which is not representable in signed 3 bits.


whitequark:
Thank you!

PR #294, opened 2020-01-05
Signal: use reset.value if that field exists.

Fatsie:
This will allow a Const or an Enum to be passed as reset value when calling Signal().

### DIFF
--- a/nmigen/hdl/ast.py
+++ b/nmigen/hdl/ast.py
@@ -797,7 +797,7 @@ class Signal(Value, DUID):
         name this ``Signal`` is assigned to. Name collisions are automatically resolved by
         prepending names of objects that contain this ``Signal`` and by appending integer
         sequences.
-    reset : int
+    reset : int or reset.value is int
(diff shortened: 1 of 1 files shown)


whitequark:
I think this is not the right approach. There are several issues:
  1. Sure, Python uses duck typing, but it doesn't mean you can use arbitrary properties of objects without considering what interface they implement. That there are both `Const.value` and `Enum.value` is a complete coincidence: they were never intended to implement the same interface. This is in contrast with something like a "file-like object", where there's a reasonably well-defined interface anyone can comply with.
  2. I'm not actually sure if handling `Const` as a reset value is the right thing to do. Right now it is clear what values are valid as reset values: literals. If we allow ASTs there, then it raises a question of whether we should allow constant expressions or even arbitrary expressions. Should `Const(123, 8)[:4]` be allowed as a reset value? Should `other_comb_sig + 10` be?

That said, allowing integral `Enum` values is obviously the right choice here. Feel free to send another PR implementing that or opening an issue and I'll implement it.

PR #295, opened 2020-01-05
hdl.ast: replace slice.end by slice.stop in ValueKey

psurply:
This patch fixes the following warning which is reported when
RTLIL is generated from a design that uses slices:

    nmigen/nmigen/hdl/ast.py:1505: DeprecationWarning: instead of `slice.end`, use `slice.stop`
        self._hash = hash((ValueKey(self.value.value), self.value.start, self.value.end))

Signed-off-by: Pierre Surply <pierre.surply@lse.epita.fr>


psurply:
Never mind, it looks like this got fixed with e18385b. Thanks!

whitequark:
Yep, I didn't merge it because I had a git stash already that became  e18385b.

PR #296, opened 2020-01-06
Signal: allow to use integral Enum for reset value.

Fatsie:
Follow-up to #294.

### DIFF
--- a/nmigen/hdl/ast.py
+++ b/nmigen/hdl/ast.py
@@ -797,7 +797,7 @@ class Signal(Value, DUID):
         name this ``Signal`` is assigned to. Name collisions are automatically resolved by
         prepending names of objects that contain this ``Signal`` and by appending integer
         sequences.
-    reset : int
+    reset : int or integral Enum
         Reset (synchronous) or default (combinatorial) value.
         When this ``Signal`` is assigned to in synchronous context and the corresponding clock
         domain is reset, the ``Signal`` assumes the given value. When this ``Signal`` is unassigned
@@ -833,6 +833,11 @@ class Signal(Value, DUID):
     def __init__(self, shape=None, *, name=None, reset=0, reset_less=False, min=None, max=None,
                  attrs=None, decoder=None, src_loc_at=0):
         super().__init__(src_loc_at=src_loc_at)
+
+        if isinstance(reset, Enum):
+            reset = reset.value
+        if not isinstance(reset, int):
+            raise TypeError("Reset value has to be an int or an integral Enum")
 
         # TODO(nmigen-0.2): move this to nmigen.compat and make it a deprecated extension
         if min is not None or max is not None:
--- a/nmigen/test/test_hdl_ast.py
+++ b/nmigen/test/test_hdl_ast.py
@@ -755,6 +755,14 @@ class SignalTestCase(FHDLTestCase):
         self.assertEqual(s1.reset, 0b111)
         self.assertEqual(s1.reset_less, True)
 
+    def test_reset_enum(self):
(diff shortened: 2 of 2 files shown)


nmigen/hdl/ast.py
```diff
@@ -834,6 +834,11 @@ def __init__(self, shape=None, *, name=None, reset=0, reset_less=False, min=None
                  attrs=None, decoder=None, src_loc_at=0):
         super().__init__(src_loc_at=src_loc_at)
 
+        if isinstance(reset, Enum):
+            reset = reset.value
+        if not type(reset) == int:
```
whitequark: This should be an `isinstance(reset, int)`, or `reset=True` will fail. (`bool` is derived from `int` in Python.)

nmigen/hdl/ast.py
```diff
@@ -797,7 +797,7 @@ class Signal(Value, DUID):
         name this ``Signal`` is assigned to. Name collisions are automatically resolved by
         prepending names of objects that contain this ``Signal`` and by appending integer
         sequences.
-    reset : int
+    reset : int, bool or integral Enum
```
whitequark: Nit: `bool` is an integral type, so it's fine to just say `int or integral Enum`; this covers the cases where it's bool or bool Enum.

whitequark:
Thanks. Could you add a test as well?

nmigen/hdl/ast.py
Fatsie: I prefer to mention ``bool`` explicitly here as it makes clear that it is a proper use of the parameter.

Fatsie:
Pushed a test.

nmigen/hdl/ast.py
whitequark: This is inconsistent with all other docstrings nMigen has, which assume that the reader knows `bool` is an integral type.

Fatsie: I'll change it but it is not how I read autodocs. To me it's not about knowing if ``bool`` is an integral type but if ``int`` as a specification in docstring means one is also supposed to pass  ``bool`` for that parameter. To me just specifying ``int`` means one is not supposed use ``bool`` for that parameter.

Fatsie:
Pushed change that removes reference to bool in docstring and extended test to check for non-integral Enum type.

nmigen/hdl/ast.py
whitequark: That's not how object-oriented programming works. If `bool` derives from `int`, then it is OK to pass `bool` anywhere you'd pass `int`; this is called the Liskov Substitution Principle. Over-explaining LSP in a docstring is like over-explaining classes in a docstring: sure, you could do it, but why?

nmigen/hdl/ast.py
```diff
@@ -834,6 +834,11 @@ def __init__(self, shape=None, *, name=None, reset=0, reset_less=False, min=None
                  attrs=None, decoder=None, src_loc_at=0):
         super().__init__(src_loc_at=src_loc_at)
 
+        if isinstance(reset, Enum):
+            reset = reset.value
+        if not isinstance(reset, int):
+            raise ValueError("Reset has to be an int or an integral Enum")
```
whitequark: This should be a `TypeError`, and talk about reset *value*, not just reset, which is a different concept.

Fatsie:
``ValueError`` -> ``TypeError``

nmigen/hdl/ast.py
Fatsie: Done.

whitequark:
Thanks for the PR!

PR #298, opened 2020-01-07
back.verilog: honour write_verilog_opts.

peteut:
`write_verilog_opts` is not properly forwarded, this should fix that.

### DIFF
--- a/nmigen/back/verilog.py
+++ b/nmigen/back/verilog.py
@@ -69,9 +69,13 @@ write_verilog -norename {write_verilog_opts}
 
 def convert_fragment(*args, strip_internal_attrs=False, **kwargs):
     rtlil_text, name_map = rtlil.convert_fragment(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


whitequark:
The current behavior is intentional. What are you trying to do?

peteut:
> The current behavior is intentional. What are you trying to do?

I tried to shut up *Vivado* by passing in some `write_verilog_opts` options. If the current behaviour is intentional we can close this request. Specifically I'm getting critical warnings *[Netlist 29-72]*, there was a hint to use *binary*, so I passed in `-nohex`, but it did not help anyway.

whitequark:
> Specifically I'm getting critical warnings _[Netlist 29-72]_, there was a hint to use _binary_, so I passed in `-nohex`, but it did not help anyway.

I agree that avoiding things like "critical warnings" from Vivado is valuable. I think this should be done by adding options to `emit_verilog` in the appropriate platform itself, not in your designs. If you manage to find the correct combinations of options for that, please do submit a PR adding them to `vendor.xilinx_*`.

PR #300, opened 2020-01-08
Intel altiobuf_ requires oe with the same with as data lines.

### LINKED ISSUE #297: Output enable for Intel platform on multi I/O pins
Not sure wether it is a bug or I have done something wrong.

I tried to get a 16 bit wide io pins, but the generated oe wire is only 1 bit. For the Intel altiobuf_bidir output enable signal must be as wide as the data lines.

Attached is a small python script to generate the output.

[ioTest.py.zip](https://github.com/m-labs/nmigen/files/4027867/ioTest.py.zip)

schwigi:
Ah ok missunderstood the concept.

This should fix #297, with your suggestion.

Simple test looks fine on MAX10 FPGA.

### DIFF
--- a/nmigen/vendor/intel.py
+++ b/nmigen/vendor/intel.py
@@ -221,8 +221,7 @@ class IntelPlatform(TemplatedPlatform):
 
     @staticmethod
     def _get_oereg(m, pin):
-        # altiobuf_ requires an output enable signal for each pin, nmigen is generating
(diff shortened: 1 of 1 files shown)


whitequark:
This is definitely the wrong fix because it has different behavior for XDR=0 and XDR>0.

whitequark:
I apologize, I misunderstood your patch here. This is, in fact, the right fix.

PR #306, opened 2020-01-17
vendor.lattice_ecp5: Support internal oscillator (OSCG)

miek:


nmigen/vendor/lattice_ecp5.py
```diff
@@ -294,15 +300,19 @@ def create_missing_domain(self, name):
         # network, its deassertion may violate a setup/hold constraint with relation to a user
         # clock. To avoid this, a GSR/SGSR instance should be driven synchronized to user clock.
         if name == "sync" and self.default_clk is not None:
-            clk_i = self.request(self.default_clk).i
+            m = Module()
+            if self.default_clk == "OSCG":
+                clk_i = Signal()
+                m.submodules += Instance("OSCG", p_DIV=128, o_OSC=clk_i)
```
whitequark: Should the frequency be configurable?

miek: Yeah, I think so. Probably best to expose as `oscg_div` (and change default_clk_constraint to use that) rather than try and calculate it?

whitequark: Sure, seems sensible to me: less chance for things to go wrong then.

whitequark:
Merged with minor stylistic modifications.

PR #324, opened 2020-04-15
use declarative setuptools config

graingert:


setup.py
```diff
@@ -1,43 +1,3 @@
...
-    return {
-        "relative_to": __file__,
-        "version_scheme": "guess-next-dev",
-        "local_scheme": local_scheme
-    }
-
-
-setup(
-    name="nmigen",
-    use_scm_version=scm_version(),
```
graingert, line 19: use_scm_version and setup_requires=["setuptools_scm"] kwargs "are considered legacy behavior"

